PR #5, opened 2013-08-07
Bugfix for the Android Cipher issue.

binarycounter:
Fixes the cipher problem with bouncycastle and the wierd java dialect of android. Thid could cause a massive flood of mineceaft related apps. ;)


binarycounter:
Major Failure! I am so sorry! Please undo my commit as it doesn't work. I don't know why, but the current code will not compile on the standard JVM, only on dalvik. Please undo my changes until i figured out what went wrong thank you. I sincerly apologize for screwing up your code.

PR #25, opened 2014-02-01
1.7.4

earthonion:
LAN chat json is different. Added getLANText() method.


earthonion:
I noticed that like an hour after pushing. Sorry about that.

PR #70, opened 2015-01-25
Fix Rotation throwing IOException in writeEntityMetadata

Johni0702:
Does what it says. It's basically just a missing break in the switch statement causing it to fall through to the default case.
Feel free to commit it yourself, imho it's not worth a merge. But do whatever's more convenient for you.


Steveice10:
Pulled, thanks!

PR #74, opened 2015-03-16
Several small additions/fixes

Johni0702:
This PR consists of five different fixes / additions.

https://github.com/Steveice10/MCProtocolLib/commit/954b799f885b43121b7be52e3feb1c3bbdc511ad
Adds several missing magic value entries and corresponding enum values and fixes a few incorrect magic values.

https://github.com/Steveice10/MCProtocolLib/commit/584b296f0bfb047cf43098fac7b76f81ec23c6f4
Adds a getter for the score type to the ´ServerScoreboardObjectivePacket´ as it is not yet accessible.

https://github.com/Steveice10/MCProtocolLib/commit/b7f071fb91b184abcc22d767c9b2e59a7dc1b1cb
Changes the first constructor of ´ServerUpdateScorePacket´ to include the objective as it's mandatory for that packet.

https://github.com/Steveice10/MCProtocolLib/commit/c912787d39763adf3c6b801be609bc0c937f3b57
Minecraft sometimes (see [here](http://minecraft.gamepedia.com/Attribute#Vanilla_Modifiers)) generates the UUID of an `AttributeModifier` instead of using one of the predefined magic values. Reading and writing back out such a packet fails as the magic value isn't found.

This commit allows for creation of `AttributeModifier`s by UUID instead of only magic value and then tries to load the magic value failing silent if it doesn't exist. Sending a packet then uses the UUID instead of the magic value allowing for unknown UUIDs to be read and written without any unexpected behavior.

https://github.com/Steveice10/MCProtocolLib/commit/48c03a93241e6a899886c4c7d121cf2b0e5f3703
Adds `equals(Object)` and `hashCode()` methods to every class for which there is a reasonable chance of it ending up in Map (basically every class which acts as a container for data). This does not include packets, simply because I cannot see them being used by anyone in a way that you'd need to have those methods implemented.
There is also the option to drop this commit and use lombok to auto-generate these methods during compile time. However something should be done as currently `new Position(0, 0, 0).equals(new Position(0, 0, 0))` returns `false` which it clearly shouldn't.

Feel free to cherry-pick whichever commit you want, I just didn't like to create five different PR as I dislike all those merge commits generated by github.


Steveice10:
Wow, lots of fixes, thanks a ton!

The missing or broken magic values are for the most part from the update process. I usually go by wiki.vg's protocol history page, which usually doesn't list things like added sounds or entities. Other times I might copy/paste a previous entry and forget to change the value of it.

PR #80, opened 2015-04-03
Shrink Imports

phase:
No need for that many lines of imports. :stuck_out_tongue_closed_eyes:


Steveice10:
I'd call that an abuse of wildcard imports. I like to be fairly specific about imports.

PR #87, opened 2015-04-05
Split window properties into separate enums for separate window types

Johni0702:
### Why is this change necessary?

This PR changes how window properties are handled which breaks backwards compatibility with programs using these classes.
However this change is necessary due to the fact that as of 1.8 there are seven different properties and e.g. up to four meanings for one of them. The alternative of having enum values named `BREW_TIME_OR_FURNACE_PROGRESS_OR_ENCHANT_1_OR_REPAIR_MAXIMUM_COST` is just ugly.
### The new `WindowProperty` structure

First of all, `WindowProperty` is now an `interface` instead of an `enum`. This is in order to allow different windows to have different properties. The different window properties themselves are `enum`s and implement the `WindowProperty` interface. They are located in the `data/game/values/window/property` package. Documentation for each of the properties is provided as their precise purpose might be hard to guess from their names.
In addition the `EnchantmentTableProperty` provides three static methods to pack/unpack an enchantment level and type into one integer as it's used by the `ENCHANTMENT_SLOT_X` properties.
### Changes to the window property packet

The `ServerWindowPropertyPacket` does not know which window type the current window has, therefore it stores the window type as a raw integer. It also provides a constructor and a getter for this value.
The main way of accessing the window property now would be to use the new getter which takes the expected type of window property and returns the appropriate property.
A constructor for creating a new packet using one of the window properties is provided as well.

### DIFF
--- a/src/main/java/org/spacehq/mc/protocol/data/game/values/MagicValues.java
+++ b/src/main/java/org/spacehq/mc/protocol/data/game/values/MagicValues.java
@@ -8,6 +8,10 @@ import org.spacehq.mc.protocol.data.game.values.setting.Difficulty;
 import org.spacehq.mc.protocol.data.game.values.statistic.Achievement;
 import org.spacehq.mc.protocol.data.game.values.statistic.GenericStatistic;
 import org.spacehq.mc.protocol.data.game.values.window.*;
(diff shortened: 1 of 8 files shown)


Steveice10:
Thanks a ton for clearing up window properties, as well as the other changes you've submitted. Thank you also for documenting it, something I've really been procrastinating on...

PR #95, opened 2015-05-16
Fix Interface Methods

devversion:
In Eclipse, by importing the Maven Project, there are a few errors by implementing the class Packet.

For example by:
- public void read(NetInput in)
- public void write(NetOutput out)
- public boolean isPriority()

These methods are marked with the @Override Annotation, but they only implementing an interface and not a class, so you have to remove the @Override Annotation.


Steveice10:
See issue ticket.

Issue #101, opened 2015-08-20
Not working after updates


ghost:
Today i found that after several new mcprotocollib snapshots my integration tests started failing. Looks like client not connection at all. Please fix this. And please, create stable releases that will not ruin other users code after each update. thx.

``` java
    @Test
    public void testClient() {
        Server server = new Server();
        InetSocketAddress address = new InetSocketAddress("localhost", 25560);
        assertTrue(server.start(address));
        assertTrue(server.isOnline());

        Session session = new Client("localhost", 25560, new MinecraftProtocol("ensirius"), new TcpSessionFactory()).getSession();
        session.connect(true);

        assertTrue(session.isConnected()); // <--- FALSE
    }
```

```
java.lang.AssertionError
```

Steveice10:
Fixed: https://github.com/Steveice10/PacketLib/commit/4c06294c883e78581a686f3aafd287aa912ce9ae

ghost:
I am sorry, but fresh snapshot not helped for me.

``` java
    public void connect() {
        MinecraftProtocol protocol = null;

        if(password != null) {
            try {
                protocol = new MinecraftProtocol(name, password, false);
            } catch (RequestException e) {
                Throwables.propagate(e);
            }
        } else {
            protocol = new MinecraftProtocol(name);
        }

        session = new Client("localhost", TestServer.getPort(), protocol, new TcpSessionFactory()).getSession();
        session.addListener(this); // <-- Already has Disconnected event.
        session.addListener(new SessionAdapter() {
            @Override
            public void disconnected(DisconnectedEvent event) {
                System.out.println("Disconnected " + event.getReason());
            }
        });
        session.connect(true);

        assertTrue("Failed connect to server", session.isConnected());
    }
```

```
java.lang.AssertionError: Failed connect to server
```

Steveice10:
Could you add a listener to the session before connecting to see if DisconnectedEvent is being called with a reason and/or error?

ghost:
Modified my snippet. No disconnectedevent called.

Steveice10:
And you are sure you're using the latest code? The included example uses isConnected in a while loop to wait for the status query to complete, and it works fine.

ghost:
Yes, i am sure. I forgot to tell, that status query test working for me.

ghost:
Looks like problem may be on my side. Because #102 passes all tests. I found that client not showing anything if server in online mode and client trying join as offline.

UPD:
[...]

PR #102, opened 2015-08-21
Added junit tests

ghost:
I suggest add travis-ci and run these tests after every commit. This will prevent problems like #101 in future. 

I created inner classes because tests other way would not be correctly fail if something going wrong. I use CountDownLatch because creating while/sleep loop is bad practic. I personally prefer use static imports to remove excess code. But for new users may be harder to understand what going on.

We can remove "example" package and redirect users to tests. But this require some modifications to show all available features.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,6 +79,12 @@
             <artifactId>mcauthlib</artifactId>
             <version>1.0</version>
         </dependency>
+        <dependency>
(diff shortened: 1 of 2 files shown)


ghost:
Looks good to me. Not sure if this would really be able to replace the example, as they have different intents behind them.

PR #103, opened 2015-09-17
Removed unnecessary parameter from constructor

ghost:


ensirius:
Also site not working

PR #106, opened 2015-09-21
Change block id&data sending.

xDiP:


Steveice10:
See other pull request.

PR #110, opened 2015-09-29
Close stream after NBT read/write.

ghost:
Bumped minimum java version to 1.7. Closing streams on 1.6 java is bad practice. All users has at least 1.7 java atm. Also maybe we can remove uncessery streams? Netty has own streams for bytebuf. But it's require pass ByteBuf.


ghost:
- That stream doesn't need closing since its just wrapper around a NetInput/NetOutput.
- I'd rather not break Java 6 compatibility over something so trivial.

PR #113, opened 2015-11-23
Fix an exception in chunk loading for full nether chunks

yawkat:


Steveice10:
Invalid, that variable is only used to compare the actual data length to the expected without skylight to check if there is skylight. Not adding skylight is intended.

yawkat:
It might be intended, but it's wrong :) below, you are comparing the actual length to the expected value - but when there is no skylight, this expected length is exactly equal to the actual length, while when there is skylight, the expected length is 2048*section_count lower than the actual length. This causes an ArrayIndexOutOfBoundsException in pass 4 when you try to copy the skylight out of the input data.

Steveice10:
The issue would then be that the check does ">=" instead of ">". I'll fix it in a moment.

Steveice10:
Should be fixed.

yawkat:
Couldn't this still happen when biome data is also present? I don't think this is robust enough.

PR #148, opened 2016-04-03
Update to MC 1.9.2

finalchild:


ghost:
It seems I have to handle
"Changed dimension in Join Game from a byte enum to an int enum (not VarInt), to the benefit of Minecraft Forge" 
this change.

ghost:
Fixed it. Tested. Successful.

PR #164, opened 2016-05-15
MC 1.9.4 support

finalchild:
I think it complete, but it may lack something.


Steveice10:
Has this been tested?

ghost:
Since I don't have any testing methods, I couldn't.
I just tracked wiki.vg's protocol history.

Steveice10:
Have you at least tried to log in to a server with it?

kukelekuuk00:
The version number still says 1.9.2

kukelekuuk00:
I get "Internal Network Exception" when trying to connect to a server (1.9.4 bungeecord) with this update.

goosewoman:
https://gist.github.com/kukelekuuk00/ddf4b5108e69ca1f3953145adb1e291c

Is what I get on my chatclient trying to connect to a 1.9.4 bungeecord server. Unsure what kind of exception is actually thrown, because it gets caught before I can touch upon it.

Might I recommend an easier way of figuring out where the error lies? Like not catching the Exceptions before I can even begin to try to figure out if they're thrown. (iirc the exception is caught in PacketLib)

ghost:
I'll try to test it today.

ghost:
I found a bug. Fixing it.

ghost:
Yup fixed
- Tested with my chat client.
  https://github.com/finalchild/MCChatClient

Issue #159, opened 2016-04-21
java.lang.IllegalArgumentException: Value 24 has no mapping for key class org.spacehq.mc.v1_9_2.protocol.data.game.entity.Effect.


MatteCarra:
http://pastebin.com/Y6wDXtRS
Happened on log into my server Mc.craftymynes.com

ghost:
#169 should fix it

PR #169, opened 2016-05-25
Fix #163 (ServerBlockValuePacket) and #159 (Effect)

### LINKED ISSUE #163: Value 0 has no mapping for key class org.spacehq.mc.protocol.data.game.world.block.value.GenericBlockValueType
I experienced this on 1.9.2 server

Full crashlog: http://pastebin.com/9LXgxNML

To reproduce, place a piston within render distance of the player and extend it

finalchild:


Steveice10:
Thanks.

PR #193, opened 2016-08-24
Check for negative GameModes in ServerPlayerListEntryPacket

phase:
These are sometimes negative for some reason. I was only able to reproduce this on one server, but this fixed it. The server was sending out a GameMode of -1.


phase:
Confirmed it to be `ADD_PLAYER`:

```
ServerPlayerListEntryPacket ADD_PLAYER GameMode: 1
ServerPlayerListEntryPacket ADD_PLAYER GameMode: -1
ServerPlayerListEntryPacket ADD_PLAYER GameMode: -1
```

I believe this to be from an IRC plugin.

Edit: Purple IRC is [sending OfflinePlayers](https://github.com/cnaude/PurpleIRC/blob/master/src/main/java/com/cnaude/purpleirc/Utilities/NetPacket_110.java) which could have a GameMode of -1.

PR #204, opened 2016-12-05
Update MobType.java

ghost:
Register ArmorStands..


ghost:
Are you sure this is a mob type? I don't see it listed on the protocol wiki.

PR #237, opened 2017-06-09
Update to MC 1.12

Johni0702:
Based on http://wiki.vg/index.php?title=Pre-release_protocol&oldid=12924

Note that this has not yet been tested **at all** because I require Minecraft Forge for my use case.
If you happen to know of any simple proxy based on MCProtocolLib that can be used to test this (e.g. one that just `.toString()`s all or some of the packets passing through), I'll happily use it to test this PR.

Recipe IDs are not passed through MagicValues and kept in their integer form instead. The reason for this is simply that atm the recipe IDs are generated automatically on game startup, depending on the load-order of the recipes. As such they are very likely to completely change if a recipe is added.
Furthermore, custom recipes are probably going to be added in 1.13 which will change all of it anyway.


Steveice10:
Can you change the pom version to "1.12-1-SNAPSHOT"? The "-1" is the MCProtocolLib revision number, changed in the event that further bugs are discovered after a specific protocol version has a release.

Other than that, this looks great. Thanks a ton.

NovaBlastHD:
Looks really good. Ran it myself, only problem I have encountered so far is that the parrot isn't registered (value 105)
Nice work!

Johni0702:
I've changed the pom version to include the revision number.

I also did some testing and fixed all bugs I could find.

The BuiltinSounds have been generated using the following command and have not been tested extensively (just a few random samples):
```bash
cd ~/.minecraft/assets
cat objects/$(cat indexes/1.12.json | jq -r '.objects."minecraft/sounds.json".hash' | sed 's/\(.\{2\}\)/\1\/\1/') | jq -r 'keys[]' | sed 's/\./_/g' | sed 's/$/,/' | tr '[:lower:]' '[:upper:]'
```

NovaBlastHD:
Thank you @Johni0702. This is very good! Works very well for 1.12 so far. Tested it yesterday, and tried to add some of the missing stuff. Now testing it today it works like a charm.

PR #283, opened 2017-09-19
Update to MC 1.12.2

Johni0702:
Based on http://wiki.vg/index.php?title=Protocol&diff=13363&oldid=13339

Also includes a commit to be tagged as the 1.12.1 release.


DaMatrix:
can confirm that this works

0-x-2-2:
same

PR #416, opened 2018-07-18
Update to MC 1.13

Johni0702:
Based on http://wiki.vg/index.php?title=Pre-release_protocol&oldid=14132

This has not been tested on live data (or actually tested at all beyond existing tests) because I require Forge for my use-case (which has not yet been updated to 1.13). So I wouldn't recommend merging just yet.

The DeclareCommands packet requires quite a bit of parsing ([see here](http://wiki.vg/Command_Data)) and I don't need it myself, so it's a stub which simply reads/writes everything into/from a byte array for now. Same with the DeclareRecipes packet.

[This](https://bugs.mojang.com/browse/MC-131684) may affect the skylight detection in `NetUtil.readColumn`. I'm not entirely sure but will probably find out once I test on live data.


Steveice10:
Great, I'll make sure to do at least some basic testing on this soon.

In regards to the skylight issue, would it be possible to detect when there should be extra byte (s) in the buffer and skip them at the end of said buffer?

ShayBox:
ETA on 1.13 support being merged? I realize it takes awhile to update because of the block name changes, but I just wanted to know a general time expected.

Steveice10:
Sorry, I completely forgot to get back to this. Will test and merge soon.

Steveice10:
As of the latest commit, I'm able to log in and idle on a fresh 1.13.1 server without any errors. Let me know if any issues come up.

PR #449, opened 2019-03-23
Added missing break in ServerPlayerListEntryPacket

MatteCarra:
This fixes PlayerListEntry when action is UPDATE_DISPLAY_NAME


Steveice10:
Good catch, thanks.

PR #466, opened 2019-07-08
Fix magic value of GlobalEntityType.LIGHTNING_BOLT

Johni0702:
This got changed during the 1.14 update. Afaict the value was always 1 (I checked vanilla 1.13.2, 1.14 and 1.14.3, not sure about snapshots) and wiki.vg doesn't show any changes either, so I don't know why it was changed to 99.
@Paulomart do you remember where you got that from?

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
@@ -358,7 +358,7 @@ public class MagicValues {
         register(PositionElement.PITCH, 3);
         register(PositionElement.YAW, 4);
 
-        register(GlobalEntityType.LIGHTNING_BOLT, 99);
+        register(GlobalEntityType.LIGHTNING_BOLT, 1);
 
         register(MobType.AREA_EFFECT_CLOUD, 0);
         register(MobType.ARMOR_STAND, 1);


Paulomart:
I was using

```
java -cp server.jar net.minecraft.data.Main --all 
```

to get the data:

```json
  "minecraft:entity_type": {
    "default": "minecraft:pig",
    "protocol_id": 5,
    "entries": {
      (...)
      "minecraft:lightning_bolt": {
        "protocol_id": 99
      },
      (...)
    }
  },
```

Johni0702:
Ah, that makes sense. That's probably what MC uses for ServerSpawnMobPacket but (for some reason) it's using a different (hard-coded) id for ServerSpawnGlobalEntityPacket.

Paulomart:
Typical mojang move; Does the 99 in `ServerSpawnGlobalEntityPacket` cause issues or does not work? (Did not test that 😞)

Johni0702:
Yeah, it's probably from way back and no one ever bothered to clean it up.
Without this fix, reading a vanilla packet results in `java.lang.IllegalArgumentException: Value 1 has no mapping for key class com.github.steveice10.mc.protocol.data.game.entity.type.GlobalEntityType
`, writing one will cause vanilla to silently ignore it.

src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
```diff
@@ -358,7 +358,7 @@
         register(PositionElement.PITCH, 3);
         register(PositionElement.YAW, 4);
 
-        register(GlobalEntityType.LIGHTNING_BOLT, 99);
+        register(GlobalEntityType.LIGHTNING_BOLT, 1);
```
Paulomart, line 361: Maybe a possible workaround would be this

```suggestion
        register(GlobalEntityType.LIGHTNING_BOLT, 1);
        register(GlobalEntityType.LIGHTNING_BOLT, 99);
        
```

But I think nobody is using the SpawnMobPacket to create lighting bolts...

Johni0702: Great, clicking "Resolve" doesn't even submit my comment, I'm not typing that again, so here's the short version:

The only valid type id for ServerSpawnGlobalEntityPacket is 1. The vanilla server never spawns anything else and the client ignores everything else.

We could add a MobType for it but, as you said, I don't think anyone uses SpawnMobPacket for lightning bolts and I'm not even sure the vanilla client would handle it well.

Paulomart: I agree.


Keep in mind that MobType also contains all ObjectTypes, see a9d83ee61b1a4ad48066c3ed4ceff0717eda6796 and ceddf672. They are registered using their respective EntityId even though they are Objects..

Maybe what we really need is a GlobalEntityType?

Johni0702: > Keep in mind that MobType also contains alls ObjectTypes, see a9d83ee and ceddf67.

Yes, I've seen several servers spawn certain objects (e.g. armor stand) using ServerSpawnMobPacket (i.e. MobType) whereas vanilla uses ServerSpawnObjectPacket (i.e. ObjectTypes). No idea why MC makes a distinction between those in the first place, I can only guess that it's for legacy reasons.

> Maybe what we really need is a GlobalEntityType?

Not sure what you're saying. That's quite literally what we have.

Paulomart: Wow I totally missed that. I thought it was about MobType. Sorry for wasting your time :facepalm:

Paulomart (approved):

PR #482, opened 2019-08-29
the "minecraft:" isn't sent by all servers

goosewoman:
Specifically 1.13.2 servers that allow 1.14 clients to join. So it crashes on connecting to server with "java.lang.IllegalArgumentException: Value smelting has no mapping for key class com.github.steveice10.mc.protocol.data.game.recipe.RecipeType" because the mapping is minecraft:smelting

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/advancement/Advancement.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/advancement/Advancement.java
@@ -12,12 +12,12 @@ import java.util.List;
 @AllArgsConstructor
 public class Advancement {
     private final @NonNull String id;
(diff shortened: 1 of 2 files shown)


Steveice10:
I don't usually support protocol hacks for joining different versions, but since the vanilla client does I suppose MCProtocolLib should accept those mappings as well.

0-x-2-2:
Yeah the official Minecraft server/client adds `minecraft:` to the start if no mod or whatever is specified.

Steveice10:
Should be fixed by https://github.com/Steveice10/MCProtocolLib/commit/c792dc022ca598b1687f7799a6332d4f6bd4a08d. Let me know if any more related problems come up.

PR #499, opened 2019-11-26
Upgrade to 1.15

Paulomart:
~~Currently not sure what to do with the extra biome data; It seems like there are now 2 fields for that? See https://wiki.vg/Pre-release_protocol#Chunk_Data~~

It seems like mojang removed the old 256 int array field, see https://wiki.vg/Talk:Pre-release_protocol#1.15_Chunk_Biome_Data_Changes

Still todo:

- [x] Update version once released
- [x] Rename "unknown" field once known


Paulomart:
Oh yes your right 🤦‍♂

Steveice10:
Thanks a bunch for this. Now that 1.15 is officially out, I should have time to test and merge around the end of the week.

Redned235:
~~Did some testing myself and did not encounter any issues.~~

Disregard that for now. I found some fairly major issues with entities (the IDs from BAT onward were shifted by 1). I will be opening a pull request to Paulomart's repository rather than opening a new pull request here so it will automatically be added to this one.

Paulomart:
Thanks @RednedEpic for fixing the entity ids, I totally forgot about them

Redned235:
I started going through some of the sound mappings and I believe the whole sound registration is wrong altogether (as well as 1.14, 1.13 and below still work in this way). 

Using [this website](https://pokechu22.github.io/Burger/1.15.html#sounds) (which was linked on wiki.vg), all the numeric IDs of the sounds can be found, which is used in various sound packets. The values within the `BuiltinSound` class are listed in alphabetical order, in which they are also [registered](https://github.com/Steveice10/MCProtocolLib/blob/master/src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java#L1191-L1195). 

After going through this website and later confirming by decompiling the Minecraft server jar ([here](https://gist.github.com/RednedEpic/6f34a30ae2d85eddfcea1a99c0293cf3) is a snippet confirming that), this is not the order they should be registered in, thus resulting them being registered with the wrong ID. For example, the `item.armor.equip_chain` numeric ID is 15 according to that website, whilst `block.bamboo.break` is what it is in MCProtocolLib.

I will be fixing this shortly as well as including all the new 1.15 sounds. A pull request will also be opened on Paulomart's repository so it can be added here.

Steveice10:
Good catch. Looking at that website for older versions, it seems 1.12 was ID'd in alphabetical order as far as I checked, but 1.13 is not.

Redned235:
Minecraft 1.15.1 released today and from the looks of it, it was just merely a protocol bump.

Steveice10:
Manually merged and updated to 1.15.1, thanks a bunch.

PR #503, opened 2019-12-31
Remove @NonNull from a nullable parameter

MatteCarra:
Slot data can be null if present is false: https://wiki.vg/Slot_Data


Steveice10:
Good catch

PR #509, opened 2020-01-21
Set proxy to null in example

Ruukas97:
Otherwise, it will use deprecated OioSocketChannel instead of non-blocking NioSocketChannel.


Ruukas97:
I spent a while trying to figure out why the ping took 1001ms. I finally tracked down the issue, and after setting to null, the ping took only 1ms.

Steveice10:
Good catch. I'll probably just change PacketLib to not use OioSocketChannel when the proxy is NO_PROXY, and I'll leave this open for now as a reminder to get that done.

Ruukas97:
Yeah, I was thinking about that, but I don't feel knowledgeable enough about Netty and the PacketLib to attempt any changes.

Steveice10:
Apologies for the delay. I've migrated the proxy code in PacketLib to Netty's ProxyHandler classes, so all connections should use NIO now.

PR #506, opened 2020-01-02
START_RAIN and STOP_RAIN is mixed up

magicus:
This breaks GeyserMC, so the Bedrock client thinks it stops raining when it starts, and vice versa.

PR #514, opened 2020-02-09
Fix flipped start and stop for rain

Redned235:
Seems like a mistake was made in commit fbd5460d7c595c05a983917e16bf048a85e80ea2 which caused rain to start on `STOP_RAIN` and stop on `START_RAIN` (See GeyserMC/Geyser#144). This PR reverts that and fixes rain.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
@@ -690,8 +690,8 @@ public class MagicValues {
         register(UpdatedTileType.BEEHIVE, 14);
 
         register(ClientNotification.INVALID_BED, 0);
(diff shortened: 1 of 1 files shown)


Steveice10:
In https://github.com/Steveice10/MCProtocolLib/pull/506, @magicus pointed out code in the official server that indicates the values should actually be as they are now. Is this not correct?

Redned235:
Apologies for the late reply here, but I did however do some testing on this and using [this code](https://gist.github.com/RednedEpic/45ed73465fd4bd3f26a79b982faec30a), it appears that when starting rain using the current code in the repository, nothing happens, and when stopping rain, it rains. However, it seems when the values are set back to what they originally were (stop rain 1, start rain 2), it works as it's supposed to.

Steveice10:
Yep, just verified that your test code works properly with the changes in this PR.

PR #526, opened 2020-04-15
Add "vanilla Forge" support

Camotoy:
This PR fixes two issues when connecting to Forge servers. Both prevent a connection to the server. Neither of these prevent connection to a normal Minecraft server (tested with Paper).

The first issue is detailed in GeyserMC/Geyser#251 - Forge sends an identifier of "minecraft:", with nothing else afterwards. Feel free to be merciless with this code as it's definitely one of those "workarounds" rather than actually implementing.

The second refers to attributes - Forge adds three attributes to a "vanilla Forge" server: swim speed, nametag distance and entity gravity. This PR adds support for these as well.


Steveice10:
Could you change the Forge-only comment on the attributed to a javadoc note for each one? Other than that, looks good.

Camotoy:
I'm not super familiar with Javadoc, so please let me know if something needs to change.

Steveice10:
I'll probably tweak the wording a bit, but looks good to me.

PR #527, opened 2020-04-17
Support for Minecraft 1.16.1

Redned235:
Updates MCProtocolLib to the latest release of Minecraft 1.16.1. This is not intended to be merged until the release of 1.16. Updates will be pushed as Mojang releases new snapshots, and hopefully by the time 1.16 is out, this will be merge-ready shortly after release.

Reference: https://wiki.vg/Pre-release_protocol

Todo:
- [x] Update version once 1.16.1 is fully released.
- [x] Find "unknown" data in packets.
- [x] Fix up the `get` and `set` methods in `FlexibleStorage` to properly work with the new padded array format.

Forge attributes will likely need to be revisited once there is a 1.16 Forge version available, however as all vanilla attributes now use the identifier/namespace format, I have updated the Forge attributes to follow that same format for the time being. If anyone runs into any bugs with this or notices something I missed, please let me know.

I have also converted the `ChatColor` enum to a normal class that holds String constants for the previous chat color types as any Hex color is now supported as of 20w17a.


Heath123:
20w20b? Wow, this is up-to-date

Redned235:
Marking this as ready for review. Please keep in mind that the `get` and `set` methods in `FlexibleStorage` still need updating to support the new padded array format. My availability over the next few days will be inconsistent, so it may a little until I'm able to get around to this, however if anyone else wants to take a stab at fixing it, please feel free to open a PR over [here](https://github.com/GeyserMC/MCProtocolLib/tree/feature/1.16) and I'll merge it 🙂.

Redned235:
Chunks are now properly functional and working thanks to @Tim203. This is now ready to be merged/reviewed.

Steveice10:
I'll get to reviewing ASAP, but that probably won't be until the weekend.

Camotoy:
Since there were no changes other than a protocol bump, I've updated this PR to 1.16.1.

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Column.java
```diff
@@ -10,16 +11,18 @@
 public class Column {
     private final int x;
     private final int z;
+    private final boolean ignoreOldData;
+    @EqualsAndHashCode.Exclude
```
Steveice10, line 15: What is the rationale for excluding chunks from equals and hashCode?

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/FlexibleStorage.java
```diff
@@ -69,14 +60,38 @@ public void set(int index, int value) {
...
-        this.data[startIndex] = this.data[startIndex] & ~(this.maxEntryValue << startBitSubIndex) | ((long) value & this.maxEntryValue) << startBitSubIndex;
-        if(startIndex != endIndex) {
-            int endBitSubIndex = 64 - startBitSubIndex;
-            this.data[endIndex] = this.data[endIndex] >>> endBitSubIndex << endBitSubIndex | ((long) value & this.maxEntryValue) >> endBitSubIndex;
+        int cellIndex = (int) (index * divideMultiply + divideAdd >> 32L >> divideShift);
+        int bitIndex = (index - cellIndex * valuesPerLong) * bitsPerEntry;
+        data[cellIndex] &= ~(maxEntryValue << bitIndex) | (value & maxEntryValue) << bitIndex;
+    }
+
+    private static final int[] MAGIC_VALUES = {
```
Steveice10, line 68: I generally put static constants at the top of the file before other fields, not the bottom.

src/main/java/com/github/steveice10/mc/protocol/data/message/EntityHoverMessage.java
```diff
@@ -0,0 +1,70 @@
...
+            this.type = type;
+            return this;
+        }
+
+        public Builder id(@NonNull String id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder name(@NonNull Message name) {
```
Steveice10, line 28: If this is optional like the field indicates, parameter should be nullable as well.

src/main/java/com/github/steveice10/mc/protocol/data/message/ItemHoverMessage.java
```diff
@@ -0,0 +1,72 @@
...
+        private String id = "";
+        private int count;
+        private JsonElement tag;
+
+        public Builder id(@NonNull String id) {
+            this.id = id;
+            return this;
+        }
+
+        public Builder count(@NonNull int count) {
```
Steveice10, line 25: int can't be null, so NonNull is not needed here.

src/main/java/com/github/steveice10/mc/protocol/data/message/ItemHoverMessage.java
```diff
@@ -0,0 +1,72 @@
...
+            this.count = count;
+            return this;
+        }
+
+        public Builder tag(@NonNull JsonElement tag) {
```
Steveice10, line 30: If this is optional as the field indicates, it should be nullable here as well.

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
```diff
@@ -64,7 +62,11 @@ public static JsonElement toJson(Message message) {
         if(message instanceof TextMessage && message.getStyle().equals(MessageStyle.DEFAULT) && message.getExtra().isEmpty()) {
             return new JsonPrimitive(((TextMessage) message).getText());
         }
+        
+        return toJsonObject(message);
+    }
 
+    public static JsonObject toJsonObject(Message message) {
```
Steveice10, line 69: Is there a reason this was split off?

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
```diff
@@ -121,6 +123,21 @@ public static JsonElement toJson(Message message) {
             } else {
                 throw new IllegalArgumentException("Unknown NBT message type in json: " + json);
             }
+        } else if(json.has("type") && json.has("id")) {
+            EntityHoverMessage.Builder builder = new EntityHoverMessage.Builder();
+            builder.type(json.get("type").getAsString());
+            builder.id(json.get("id").getAsString());
+            if (json.has("name"))
```
Steveice10, line 130: Please explicitly use brackets. Ditto for the two occurrences below.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/ServerChatPacket.java
```diff
@@ -33,19 +36,29 @@ public ServerChatPacket(@NonNull Message message) {
...
     public ServerChatPacket(@NonNull String text, @NonNull MessageType type) {
-        this(MessageSerializer.fromString(text), type);
+        this(MessageSerializer.fromString(text), type, new UUID(0, 0));
+    }
+
+    public ServerChatPacket(@NonNull Message message, @NonNull MessageType type) {
+        this(message, type, new UUID(0, 0));
+    }
+
+    public ServerChatPacket(@NonNull String text, @NonNull MessageType type, UUID uuid) {
```
Steveice10, line 46: UUID should be NonNull here.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/ServerRespawnPacket.java
```diff
@@ -20,25 +19,37 @@
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 @AllArgsConstructor
 public class ServerRespawnPacket implements Packet {
-    private int dimension;
+    private String dimension;
```
Steveice10, line 22: Should this and worldName be NonNull? If not then this is fine.

src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
```diff
@@ -268,21 +267,13 @@
         register(GameMode.CREATIVE, 1);
         register(GameMode.ADVENTURE, 2);
         register(GameMode.SPECTATOR, 3);
+        register(GameMode.UNDEFINED, 255); // https://bugs.mojang.com/browse/MC-189885
```
Steveice10, line 270: I'm not aware yet exactly how this new undefined game mode works, but if it's only for the case where previous game mode is undefined in packets, would it make more sense to just make that field nullable to avoid confusion in the possible values of the current game mode?

Steveice10:
Looks very good, went through and left comments where I noticed anything. One thing I noticed that I couldn't seem to leave a comment for is that an empty "src/main/java/com/github/steveice10/mc/protocol/data/message/MessageStyle.java" file was created, which should be removed.

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
rtm516: If you need to get a JsonObject all the time instead of sometimes getting a JsonPrimitive this allows for that, as you cant construct it yourself because the methods used are private.

Steveice10: Yes, but I don't understand what situation you'd specifically need a JsonObject in over the simpler representation.

rtm516: It was for getting around the parent style issue, see here: https://github.com/GeyserMC/Geyser/blob/feature/1.16/connector/src/main/java/org/geysermc/connector/utils/MessageUtils.java#L208-L238

src/main/java/com/github/steveice10/mc/protocol/data/message/EntityHoverMessage.java
rtm516: Fixed

src/main/java/com/github/steveice10/mc/protocol/data/message/ItemHoverMessage.java
rtm516: Fixed

src/main/java/com/github/steveice10/mc/protocol/data/message/ItemHoverMessage.java
rtm516: Fixed

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
rtm516: Fixed

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
Steveice10: Instead of messing with the raw JSON, it would probably be a better solution if I added a method for copying an existing message into a builder, so you could just do something like:
```
message.toBuilder()
    .style(message.getStyle().toBuilder()
                 .color(parent.getStyle().getColor())
                 .formats(parent.getStyle().getFormats()))
    .build()
```

I could get that put together after the PR is merged if you want.

rtm516: That would be great, but would it be possible to get that PRd onto this PR? So I can just Geyser to work with that before this is merged to prevent any delays releasing.

Steveice10: I've committed it to the main repository, so you can go ahead and pull it into your branch.

src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
Camotoy: Vanilla behavior sends 255/-1 whenever the gamemode is undefined (usually for the new previous gamemode option). The alternative would be to do a check for gamemode 255 and then set it as null.

Steveice10: I think it would be a good idea to just allow previous game mode in particular to be null to avoid adding a confusing extra value for a special case.

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Column.java
Camotoy, line 15: I think tests were failing for no reason without that - tagging @Tim203

src/main/java/com/github/steveice10/mc/protocol/data/message/MessageSerializer.java
rtm516: This should be fixed now

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/FlexibleStorage.java
Tim203: I generally also do that, but I have to agree with Redned on this one because it is actually quite long.

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Column.java
Tim203: Yes, even though the chunks look correct.

Steveice10: Okay, I'll let it through for now but probably revisit the issue myself later.

src/main/java/com/github/steveice10/mc/protocol/data/MagicValues.java
Camotoy: Shall I add a GameMode read/write option, then, that makes it null if it detects 255?

Steveice10: Yeah, it would be a good idea to just check if it's 255 and use null specifically for optional previous game mode fields.

Camotoy: Fixed.

Camotoy:
All requests should have been fulfilled. I've also tested the current Forge 1.16 branch and removed/updated their attributes where needed.

Steveice10:
Thanks, I’ll do some final review and testing over the next few days.

Steveice10:
All merged, thanks.

Camotoy:
On behalf of the Geyser team, thank you as well!

MatteCarra:
@RednedEpic  Where are these values found? https://github.com/GeyserMC/MCProtocolLib/blob/3f5f9797e6da511284d3380ac8eef2aa3046aa0f/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/FlexibleStorage.java#L90

I can't find them in the provided link given for reference.
Thanks in advance.

PR #538, opened 2020-05-14
Added shulker box UpdatedTileType to prevent errors on older versions

rtm516:
Reference can be found here: https://wiki.vg/index.php?title=Protocol&oldid=14204#Update_Block_Entity


Redned235:
To provide a bit more context on this: despite the fact that a notchian server in 1.15.2 will not send a block entity update for shulker boxes, if you utilize the client functionality of MCProtocolLib and join a server running on 1.12 using ViaVersion, (which is a plugin that allows 1.15.2 clients to join older servers), the server on that version still sends a block entity update for shulker boxes, and thus will print an error like this:
81801051-8028c580-951c-11ea-9efc-51cb0b9d4d00

PR #553, opened 2020-06-18
Remove BlockState class

Camotoy:
The BlockState class is simply a wrapper for an integer. Dealing with many BlockState classes in our Geyser project uses up a lot of RAM when we can simplify that down to more efficient integers.

Definitely open to discussion.


Steveice10:
Thanks, this class was a hold-over from when there were distinct ID and data components, and doesn't really make much sense anymore.

Camotoy:
:+1:

PR #562, opened 2020-07-10
 Fix "set" in FlexibleStorage

Heath123:
The assignment using `&=` seems to make it so that once a value is set to 0 it can never be set to anything else, because when you do a bitwise AND with 0 it stays as 0. This means that you can't write data to chunks, which use FlexibleStorage. Setting this to `=` fixes that. Is there any reason why this was `&=` or was it a mistake?
Edit: That didn't work, @Johni0702 explained below, but it should be right now

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/FlexibleStorage.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/FlexibleStorage.java
@@ -84,7 +84,7 @@ public class FlexibleStorage {
 
         int cellIndex = (int) (index * divideMultiply + divideAdd >> 32L >> divideShift);
         int bitIndex = (index - cellIndex * valuesPerLong) * bitsPerEntry;
-        data[cellIndex] &= ~(maxEntryValue << bitIndex) | (value & maxEntryValue) << bitIndex;
+        data[cellIndex] = ~(maxEntryValue << bitIndex) | (value & maxEntryValue) << bitIndex;
     }
 
     public FlexibleStorage transferData(int newBitsPerEntry, IntFunction<Integer> valueGetter) {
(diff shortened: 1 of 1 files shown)


Johni0702:
(from https://github.com/Steveice10/MCProtocolLib/commit/7545884a2dd0b89ee93f4d9ca98b1b8321681847#r40504585)

> Setting this to just `=` doesn't seem right either, that'll overwrite all other values in the cell.
> What it looks like to me is that the author tried to simplify the original line (76) but got the operator precedence wrong:
> 
> ```
> // Original line:
> this.data[startIndex] = this.data[startIndex] & ~(this.maxEntryValue << startBitSubIndex) | ((long) value & this.maxEntryValue) << startBitSubIndex;
> // They read:
> this.data[startIndex] = this.data[startIndex] & (~(this.maxEntryValue << startBitSubIndex) | ((long) value & this.maxEntryValue) << startBitSubIndex);
> // But it actually means:
> this.data[startIndex] = (this.data[startIndex] & ~(this.maxEntryValue << startBitSubIndex)) | ((long) value & this.maxEntryValue) << startBitSubIndex;
> ```

Heath123:
Oh, the build failed... Hang on, I'll fix it

Steveice10:
@DoctorMacc Does this resolve the chunk issue you were experiencing?

Camotoy:
I will check now.

Heath123:
It resolved it for me, so yes, it should

Camotoy:
Yes, I can confirm that this fixes my issue, and the `@EqualsAndHasCode.Ignore` in Column can be removed and tests pass successfully. Well done, Health and Johni.

Steveice10:
Great, I’ll merge this then and put out a commit later today removing the ignore.

Tim203:
My bad. Glad to see that it is fixed

PR #578, opened 2020-09-26
Add an option to disable the automatic keep alive packet options

Camotoy:
That way the clients/servers can more finely control the keep alive process.


Steveice10:
I would prefer if this used a client/server flag similar to compression threshold and other settings, rather than a field in MinecraftProtocol.

Camotoy:
Implemented as such.

PR #588, opened 2020-10-14
Palette optimization

DaMatrix:
This eliminates a lot of unnecessary boxing when reading chunk data by making `ListPalette` and `MapPalette` use an `int[]` instead of a `List<Integer>` and `Map<Integer, Integer>`, respectively.

`MapPalette#stateToIds` now uses an `IntObjectMap<Integer>` instead of `Map<Integer, Integer>`. This is significantly better because it eliminates boxing allocations for keys (which will account for the majority of boxing overhead), and the values only have to be boxed once.

`Chunk#resizePalette` no longer copies all the old data into an `int[]` before transferring the data into the resized `BitStorage`. There seems to be no reason for this, as far as I can tell it's just needless allocation of a rather large array.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Chunk.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Chunk.java
@@ -99,14 +99,14 @@ public class Chunk {
 
     private void resizePalette() {
         Palette oldPalette = this.palette;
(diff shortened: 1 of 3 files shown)


Steveice10:
Thanks for this.

The int[] for old data was an attempt to avoid recomputing cell and bit indexes for every single block when transferring between storage instance, but ultimately I suppose it would have only given marginal gains if any, and this is much cleaner.

PR #592, opened 2020-10-21
Add full support for the UNKNOWN gamemode

Camotoy:
Previously we correctly handled the unknown gamemode for 'previous gamemode'; however, this functionality can extend to the main gamemode. The gamemode writing we've been doing has been replaced with the UNKNOWN enum, so we aren't dealing with a null value for parameters that are marked NonNull.


Steveice10:
Out of curiosity, in what kind of situation is current game mode unknown?

Camotoy:
It appears that more complicated setups (such as minigames) can take advantage of it - https://github.com/GeyserMC/Geyser/issues/1362 is where I'm coming from.

PR #597, opened 2020-11-02
Update to 1.16.4

Camotoy:
Only protocol change is the book edit packet.


Steveice10:
Manually merged so I could tag a 1.16.3 release first, thanks.

PR #598, opened 2020-11-04
[WIP] Support for Minecraft 1.17

Camotoy:
Updates MCProtocolLib to 1.17.0.


Camotoy:
@Steveice10 I've consolidated the initClientGame and initServerGame functions in MinecraftProtocol so we're only registering packets in one function. This way, it's far easier to add new packets to the function without shifting two sets of IDs. I recommend you take a look and see if you like it.

Steveice10:
Yeah, looks good to me, thanks.

CrushedPixel:
@Camotoy can you make sure to add the `@With` annotation to any newly added packets, in accordance with #620 ?

Camotoy:
Sure.

Camotoy:
This PR has been updated for 21w17a. Of note - I'm not sure on how to handle the UpdateLight packet, so currently the variables match Minecraft's own packet. Feel free to PR to the repository and alert me here if you wish to change it.

Camotoy:
As of 21w19a, Minecraft is going to target Java 16, for information.

Astralchroma:
This needs to be ready soon as Minecraft Java 1.17 will be releasing June 8th.

Source: https://www.minecraft.net/en-us/article/caves---cliffs--part-i-has-release-date-

Camotoy:
This will be usable by then for sure. Sound definitions still need to be updated and the sculk sensor packet needs to be finished.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/entity/player/ServerPlayerCombatKillPacket.java
```diff
@@ -0,0 +1,40 @@
...
+
+@Data
+@With
+@Setter(AccessLevel.NONE)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class ServerPlayerCombatKillPacket implements Packet {
+    private int playerId;
+    private int killerId;
+    private Component message;
```
Steveice10, line 20: This looks like it should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/title/ServerSetActionBarTextPacket.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+
+@Data
+@With
+@Setter(AccessLevel.NONE)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class ServerSetActionBarTextPacket implements Packet {
+    private Component text;
```
Steveice10, line 18: This looks like it should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/title/ServerSetSubtitleTextPacket.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+
+@Data
+@With
+@Setter(AccessLevel.NONE)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class ServerSetSubtitleTextPacket implements Packet {
+    private Component text;
```
Steveice10, line 18: This looks like it should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/title/ServerSetTitleTextPacket.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.io.IOException;
+
+@Data
+@With
+@Setter(AccessLevel.NONE)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class ServerSetTitleTextPacket implements Packet {
+    private Component text;
```
Steveice10, line 18: This looks like it should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/world/ServerAddVibrationSignalPacket.java
```diff
@@ -0,0 +1,42 @@
...
+
+//TODO
+@Data
+@With
+@Setter(AccessLevel.NONE)
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor
+public class ServerAddVibrationSignalPacket implements Packet {
+    private Position origin;
+    private String identifier;
```
Steveice10, line 19: I see this packet is still a WIP, but preliminarily, origin and identifier look like they should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/world/ServerUpdateLightPacket.java
```diff
@@ -1,48 +1,50 @@
...
-    private static final NibbleArray3d EMPTY = new NibbleArray3d(EMPTY_SIZE);
-
     private int x;
     private int z;
+    private BitSet skyYMask;
+    private BitSet blockYMask;
+    private BitSet emptySkyYMask;
+    private BitSet emptyBlockYMask;
+    private List<byte[]> skyUpdates;
+    private List<byte[]> blockUpdates;
```
Steveice10, line 25: These all look like they should be NonNull.

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/Column.java
```diff
@@ -13,17 +13,13 @@
     private final @NonNull Chunk[] chunks;
     private final @NonNull CompoundTag[] tileEntities;
     private final @NonNull CompoundTag heightMaps;
-    private final int[] biomeData;
+    private final @NonNull int[] biomeData;
 
     public Column(int x, int z, @NonNull Chunk[] chunks, @NonNull CompoundTag[] tileEntities, @NonNull CompoundTag heightMaps) {
-        this(x, z, chunks, tileEntities, heightMaps, null);
+        this(x, z, chunks, tileEntities, heightMaps, new int[0]);
```
Steveice10, line 19: According to the wiki this should be size 1024, not 0. In fact this constructor should probably be at least deprecated since biome data is now required.

ghost (commented):
<removed>

Camotoy:
On the non-null components: this needs to be looked into because there is an existing bug in the 1.16.5 code where the server sends null for the ServerTitlePacket:

```
Downstream packet error! Don't know how to turn null into a Component
[15:23:53] [nioEventLoopGroup-4-2/ERROR]: com.google.gson.JsonParseException: Don't know how to turn null into a Component
[15:23:53] [nioEventLoopGroup-4-2/ERROR]:     at org.geysermc.platform.bungeecord.shaded.adventure.text.serializer.gson.ComponentSerializerImpl.notSureHowToDeserialize(ComponentSerializerImpl.java:240)
[15:23:53] [nioEventLoopGroup-4-2/ERROR]:     at org.geysermc.platform.bungeecord.shaded.adventure.text.serializer.gson.GsonComponentSerializerImpl.deserialize(GsonComponentSerializerImpl.java:84)
[15:23:53] [nioEventLoopGroup-4-2/ERROR]:     at org.geysermc.platform.bungeecord.shaded.adventure.text.serializer.gson.GsonComponentSerializerImpl.deserialize(GsonComponentSerializerImpl.java:40)
[15:23:53] [nioEventLoopGroup-4-2/ERROR]:     at com.github.steveice10.mc.protocol.packet.ingame.server.ServerTitlePacket.read(ServerTitlePacket.java:62)
```

Implying that it can be null? I have yet to find a way to reliably replicate that, and if every new packet is affected.

Steveice10:
Interesting, okay. If this is intended behavior we may need some special handling of the "null" string to allow for null components in places where they are possible.

Camotoy:
The tests keep failing because the chunk array size is not dependent on the BitSet size, but rather on the world height given in the JoinGame or Respawn packets. Java Edition itself does not read the chunk data in its packet class. Therefore the proper solution (from my thought process) is to abstract out the chunk data and have the implementations read and write using their stored data; MCProtocolLib itself only reads and write a byte array. If there's a better way to handle it, let me know.

Steveice10:
That would probably be the most proper way to do it. However, it might also be fine to give the test custom comparison logic that only compares up to the shortest array length and then makes sure the tail end of the longer array is all nulls. Realistically, having a trail of nulls at the end of an incoming packet's chunks should not be an issue since the bits for them will always be 0 anyway.

Camotoy:
Tests are fixed, but it appears that this is a prime example of bandaid-ing a broken test. I just ensured that the length was an amount that would transfer correctly.

Steveice10:
As long as the packet writes and reads the data correctly, that's what the test sets out to check. Any number of extra nulls at the end of the array has no real impact because they won't be marked in the bit mask. If you want to implement a solution that preserves the exact data structure of the packet (e.g. lazy parsing of the chunk data byte array) that's fine too.

Camotoy:
Superseded by #632.

PR #605, opened 2020-11-21
Add a raw message attribute to ServerChatPacket

rtm516:
This allows projects such as Geyser to get the raw json string from the message instead of converting the parsed message back. This means we don't have to worry about slight conversion inconsistency, a better solution would be to replace the current message class and handling with a library such as [Adventure](https://github.com/KyoriPowered/adventure) but that's out of scope for this PR.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/ServerChatPacket.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server/ServerChatPacket.java
@@ -4,7 +4,6 @@ import com.github.steveice10.mc.protocol.data.MagicValues;
 import com.github.steveice10.mc.protocol.data.game.MessageType;
 import com.github.steveice10.mc.protocol.data.message.Message;
 import com.github.steveice10.mc.protocol.data.message.MessageSerializer;
-import com.github.steveice10.mc.protocol.data.message.MessageSerializer;
 import com.github.steveice10.packetlib.io.NetInput;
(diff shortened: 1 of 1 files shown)


Steveice10:
Looking at this, I think you would still end up with a constructor that takes both Message and a raw message String, which doesn't really make sense.

In general, I think it would be better to just only store the raw string in the packet class. Then it could have a constructor variants for both Message and String as well as a getMessage that parses the string on demand instead of at every read.

Steveice10:
As a side note on the topic of Adventure, I can't say I have any experience with it but just from a cursory look it seems to have a lot of extra complexity that just wouldn't make sense to bundle in with MCProtocolLib.

rtm516:
We would only need to use the `adventure-text-serializer-gson` module and the core module for json to component creation and then just use the `Component` class instead of `Message`. For the most part it should be an easy change and mean this lib doesn't have to change chat stuff when it gets updated.

Steveice10:
Makes sense. Switched over to Adventure in https://github.com/Steveice10/MCProtocolLib/commit/ccbd52dffd343be7f42b2b9545b90e4465963396. Feel free to open a new issue/PR if there's anything else that you need.

Steveice10:
Took a closer look at your use-case and realized you might need to be able to swap out the serializer, so I added that capability in https://github.com/Steveice10/MCProtocolLib/commit/26201a49f082e49b55ee592b2a52fecc6da79ebe.

rtm516:
Wow thank you, working on updating Geyser with these changes.

PR #612, opened 2020-12-16
Removed client token from MinecraftProtocol

MatteCarra:
Client token is NOT part of the Minecraft protocol.
In fact it's only used by the Mojang authentication that is going to be discontinued soon.
This change is required for: https://github.com/GeyserMC/MCAuthLib/pull/1


Steveice10:
Going to hold off on merge to see how the linked PR ends up first.

MatteCarra:
Yes, that was the idea.
Btw I fixed a typo in the first comment

Steveice10:
MCAuthLib changes are still pending, but I'm going to go ahead and merge this since I think it would be a good idea to start encouraging doing authentication outside of MinecraftProtocol.

PR #614, opened 2021-01-14
Fix ServerWorldBorderPacket initialization

Camotoy:
Seems like these two were flipped around.


Steveice10:
Could you also swap them around in write() to match?

Camotoy:
Whoops, thanks for the catch! Addressed.

PR #640, opened 2021-08-09
Change Constant from -MAX_VALUE to MIN_VALUE

Newspicel:


Steveice10:
This does not seem to be correct. According to the protocol wiki Minecraft uses -MAX_VALUE: https://wiki.vg/Command_Data#Properties

PR #662, opened 2021-10-06
Refactor to use mojmap packet names and bump version

basaigh:


NathanKassab:
Personally I don't think this will do anything but cause confusion for people who have used the lib before and problems for people who are updating their older projects to newer versions

Camotoy:
I think the net benefit of having a unified naming system rather than people having to guess if a certain packet matches documentation/another project is more beneficial. However, you raise a good point - I think we could easily have a migration guide for projects to update against.

Camotoy:
I'm going to merge this in for 1.18 and set up a migration guide similar in format to https://gist.github.com/KennyTV/3778685985ddc45bd99e003191ff02eb.

PR #665, opened 2021-10-20
Updated Lombok to fix broken compilation when using Java 16

tycrek:
I was unable to compile using JDK 16 due to an issue with Lombok. [This StackOverflow answer](https://stackoverflow.com/a/66981165/9665770) describes the solution. Further details in https://github.com/projectlombok/lombok/issues/2681 (linked in the answer).


Camotoy:
Thanks!

PR #668, opened 2021-11-06
Add missing Generic Statistics

davchoo:


Camotoy (approved):

Camotoy:
Will merge as-is for completion's sake, but for 1.18 I want to move away from using MagicValues where possible. Looks like this all uses enums, so we can just check enum ordinals instead of using the MagicValues system.

PR #681, opened 2022-01-05
Fix resizing a singleton palette

MWHunter:
The current method accomplishes nothing because everything is already set to 0 in the data array.  I've replaced it with a method that sets the ID of 0 to point to whatever the singleton palette was pointing at before.

Implemented and tested with the nearly identical system PacketEvents 2.0 to be working - as I've copied and pasted over many of these classes.

Without this fix, the first block that is set in this resized singleton palette becomes the identity of every single block in that chunk section - which is quite bad.


Redned235:
Thanks for the contribution!

PR #682, opened 2022-01-11
Fixed: Added missing MoveToHotbarAction value for the swap with off-hand inventory action.

blablubbabc:
This inventory action is triggered when a player has their inventory open and clicks on some slot with the `F` key.
This case is also documented here: https://wiki.vg/Protocol#Click_Window

I noticed this because client sessions in a simple private Minecraft server proxy that uses this library were silently (without stack trace or anything) closed whenever the player uses this inventory action. Adding this MoveToHotbarAction value was confirmed to resolve the issue.


Camotoy (approved):
Thanks!

PR #683, opened 2022-01-17
Consolidate all documentation PR's (#666, #667, #673)

tycrek:
This PR combines #666, #667, #673 (while fixing errors in one of them) with some additional changes of my own. Two of the PR's are essentially identical and one of them somehow has conflicts with current `master`.

# But why another PR

[What's the solution to too many standards? Write a new standard.](https://www.explainxkcd.com/wiki/index.php/927:_Standards)


rtm516:
The maven repo should be changed to use the open collab repo as the project builds are pushed into there. See this page https://github.com/GeyserMC/Geyser/wiki/Using-Geyser-or-Floodgate-as-a-dependency

tycrek:
> The maven repo should be changed to use the open collab repo as the project builds are pushed into there.

Fixed in 8ac1923

Redned235 (approved):

PR #688, opened 2022-02-19
Make title in ClientboundOpenScreenPacket a component

mcmdev:
This should be a component and not a string. Atleast it's this way in vanilla and paper.


src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/inventory/ClientboundOpenScreenPacket.java
```diff
@@ -18,18 +20,18 @@
 public class ClientboundOpenScreenPacket implements Packet {
     private final int containerId;
     private final @NonNull ContainerType type;
-    private final @NonNull String name;
+    private final @NonNull Component name;
```
Redned235: This is going to break ABI. I suggest renaming this parameter to `title` and adding (deprecated) getName and setName methods using strings so code currently expecting this to be a string does not break.

Redned235 (changes requested):

Camotoy (approved):

Camotoy:
Thanks!

PR #695, opened 2022-04-12
Update README due to incorrect dependency examples

viktorhrekh:
The MCProtocolLib dependency requires JitPack repository as well. I've got confused why some dependencies wasn't resolved after setting up everything like in the examples. Imho there's required update to examples. Also I've added a new example for Gradle Kotlin DSL to follow the technologies trend.


viktorhrekh:
@Camotoy check it out (?)

Camotoy:
Hey, sorry for the delay here.

I think the better route to take would be to just publish MCAuthLib and OpenNBT under OpenCollab. I've already set up the necessary requirements for jars to be published; we just need that to link up to MCProtocolLib.

PR #710, opened 2022-10-16
Update GitHub Actions + Jenkins.

Kichura:
Same and last PR as [Geyser's #3351](https://github.com/GeyserMC/Geyser/pull/3351) - This follows the same pattern as the GeyserConnect PR i made, however instead of JDK 17 outright; it will be JDK 11 instead for now.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -5,13 +5,14 @@ on: [push, pull_request]
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
-      - uses: actions/checkout@v2
-      - name: Set up JDK 1.8
-        uses: actions/setup-java@v1
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


NathanKassab:
Why would you switch it to java 11? As shown in the pom this project is made for java 8

Kichura:
> Why would you switch it to java 11? As shown in the pom this project is made for java 8

I switched this to JDK 11 to see if it was a better option for compatibility reasons - if it manages to break anything in process then i can undo it.

NathanKassab:
Well in that case I don't see the point in updating, there doesn't seem to be any compatibility issues so why potentially break projects relying on this by forcing users onto a different java version?

Kichura:
"potentionally break projects" - It never was my goal to achieve nor purposely do since it would sound like JDK 11 never had any compatibility system(s) to begin with when it's the opposite by default,

I am however aware that JDK 16/17 will break things that use old code so this is one reason why i went for 11 instead so that it isn't too bleeding edge.

PR #730, opened 2023-03-21
Fix rain when set by GameEvent directly

### LINKED ISSUE #3611: Weather all time
### Describe the bug

It's raining on my bedrock client even when I use /weather clear. Fix this please, on Java it does not happens. (I am on a Java Client and the weather is good, no raining, but in Bedrock Client, it's raining every time.

### To Reproduce

1. Join Bedrock Client
2. Put /weather clear
3. Look that it's raining

### Expected behaviour

That it didn't rain

### Screenshots / Videos

_No response_

### Server Version and Plugins

image

### Geyser Dump

https://dump.geysermc.org/NNcefCdz39E7gSVqzjh9HPA9kGIB1tvu

### Geyser Version

Geyser 2.1.0-SNAPSHOT (git-master-6ec7326)

### Minecraft: Bedrock Edition Device/Version

All (I tried Windows 11)

### Additional Context

_No response_

Kas-tle:
Should fix https://github.com/GeyserMC/Geyser/issues/3611

As documented on the protocol wiki, stop rain (1) should come before start rain (2) https://wiki.vg/Protocol#Game_Event. Initially I thought that this could be related to our handling of rain strength (7) (see https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaGameEventTranslator.java#L72-L87). In fact, that code was working so well that it was masking this issue entirely.

For context, this issue was primarily experienced when the weather lock flag was set to clear with world guard. In this scenario, only a single Game Event packet is sent to set the weather to clear. But given we had these switched, it would instead lead to endless rain for the Bedrock Client.

In contrast, the Vanilla weather command gave the appearance of working fine. It only appeared this way, however, because the rain strength logic effectively ends up overriding the initial packet sent that indicates clear weather to the bedrock client. These are sent because vanilla will send multiple rain strength packets that ramp up in strength, as well as ramp down in strength when the weather is cleared.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/level/notify/GameEvent.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/level/notify/GameEvent.java
@@ -2,8 +2,8 @@ package com.github.steveice10.mc.protocol.data.game.level.notify;
 
 public enum GameEvent {
     INVALID_BED,
-    START_RAIN,
     STOP_RAIN,
+    START_RAIN,
     CHANGE_GAMEMODE,
     ENTER_CREDITS,
     DEMO_MESSAGE,


Camotoy (changes requested):
I don't think this is the right solution. I took a look through the Minecraft source, and in ClientboundGameEventPacket, START_RAIN is first.

Kas-tle:
The Game Event packet sent on `/weather clear` has an id of 1, while `/weather rain` has an id of 2. Isn't the id dictated by the order we set the enum here?

Edit: I see what you mean in the mapped jar... both Yarn and Mojang's mappings indeed do use 1 for START_RAIN and 2 for STOP_RAIN, but I am not really sure how else to explain what is actually sent to the client when rain is started versus stopped... it seems to be in total contradiction to what we have in the mappings.

Kas-tle:
Ok I think I've figured out what is going on here.

If we look in ClientPacketListener in the mojang mapped jar, we see:

```java
else if (type == ClientboundGameEventPacket.START_RAINING) {
            this.level.getLevelData().setRaining(true);
            this.level.setRainLevel(0.0f);
        } else if (type == ClientboundGameEventPacket.STOP_RAINING) {
            this.level.getLevelData().setRaining(false);
            this.level.setRainLevel(1.0f);
        }
```

So... when the client gets START_RAINING, it actually sets the rain level to 0... when it gets STOP_RAINING it sets the rain level to 1... effectively reversing what the logical naming of this would mean...

So I guess it really comes down to semantics. We can either treat their mistake as correct and invert these in Geyser, or we can take the approach of reversing them here so the names actually make sense and just comment that we have done so.

It seems they came to this same conclusion in the Paper discord: https://canary.discord.com/channels/289587909051416579/555462289851940864/1066095359404802161

basaigh:
`if (this.levelData.isRaining()) {
         this.rainLevel += 0.01F;
     } else {
         this.rainLevel -= 0.01F;
 }
 this.rainLevel = Mth.clamp(this.rainLevel, 0.0F, 1.0F);`

It looks to me as if the rain level is set like that to allow for a gradual fade-out of the rain.
(Sorry for the code formatting, was done on mobile)

Kas-tle:
Ugh so that might be the intent, but at least spigot seems to set it at the end of the rain storm... like it will ramp down the rain level first, and more or less send """start rain""" (Mojang's words) and rain level 0 at the same time. I think this confusion probably explains a lot of the weird rain problems we see that we can't really track down... there is probably a split of developers that treat this as something you send when the rain starts versus when it ends.

But at the end of the day, in a server environment, the """start rain""" packet alone will actually set the rain level to 0, causing the rain to clear, and the """stop rain""" packet alone will set the rain level to 1, causing it to rain.

Kas-tle:
Closing in favor of https://github.com/GeyserMC/Geyser/pull/3637
Still might be a good idea to add a short comment about this somewhat non-intuitive naming in the official mappings.

Issue #699, opened 2022-06-29
nioEventLoopGroup thread prevents JVM from exiting


Luluno01:
Hi, I had a similar problem with #408 with both 1.18.2-1 and 1.18.2-2-SNAPSHOT. Not sure what I did wrong but here is what happened:

The dependencies in my `build.gradle`:

```
implementation("com.github.GeyserMC:opennbt:1.4")
implementation("com.github.steveice10:packetlib:2.1")
implementation("com.github.GeyserMC:mcauthlib:6f3d6aada5")
implementation("com.github.steveice10:mcprotocollib:1.18.2-1")  // Same with 1.18.2-2-SNAPSHOT
```

Code snippet:

```Kotlin
val threadsBefore = Thread.getAllStackTraces().keys

val protocol = MinecraftProtocol("whosyourdaddy")
val sessionService = SessionService()
val client = TcpClientSession("localhost", 25565, protocol, null)
client.setFlag(MinecraftConstants.SESSION_SERVICE_KEY, sessionService)
client.addListener(object: SessionAdapter() {
    override fun packetReceived(session: Session, packet: Packet) {
        if (packet is ClientboundLoginPacket) {
            session.send(ServerboundChatPacket("Dummy client connected!"))
        } else if (packet is ClientboundChatPacket) {
            println("Received chat message: ${packet.message}")
            println("Session alive: ${session.isConnected}")
            session.disconnect("Done")
        }
    }

    override fun disconnected(event: DisconnectedEvent) {
        println("Client disconnected: ${event.reason}")
        event.cause?.printStackTrace()
    }
})
client.connect()
delay(2000)
val threadsAfter = Thread.getAllStackTraces().keys
delay(1000)
println(threadsAfter.subtract(threadsBefore))
```

And the output:

```
[...]

PR #744, opened 2023-08-03
expose the shutdown method of the event loop group

petersv5:
To exit normally all non-daemon threads of the java process are supposed to have stopped. TcpClientSession creates an static class member EventLoopGroup which in turn instantiates worker threads. These need to be stopped as a part of shutting down the server.

This patch exposes the shutdown of the EventLoopGroup as a static function intended to be called by some central function as a part of server shutdwon.

### DIFF
--- a/src/main/java/com/github/steveice10/packetlib/tcp/TcpClientSession.java
+++ b/src/main/java/com/github/steveice10/packetlib/tcp/TcpClientSession.java
@@ -316,6 +316,20 @@ public class TcpClientSession extends TcpSession {
                 break;
         }
 
-        Runtime.getRuntime().addShutdownHook(new Thread(() -> EVENT_LOOP_GROUP.shutdownGracefully()));
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> shutdownEventLoopGroup()));
+    }
+
+    public static void shutdownEventLoopGroup() {
+        if (EVENT_LOOP_GROUP != null) {
+            try {
+                EVENT_LOOP_GROUP.shutdownGracefully().sync();
+                EVENT_LOOP_GROUP = null;
+                CHANNEL_CLASS = null;
+                DATAGRAM_CHANNEL_CLASS = null;
+            } catch (InterruptedException e) {
+               e.printStackTrace();
+            }
+
+        }
     }
 }


petersv5:
This will be accompanied by a PR to Geyser using the exposed interface.

An anternative would be to insted supply the
EventLoopGroup from the outside, where it can be tracked. 

Note that the existing code shutting down the EventLoopGroup in a
process shutdown hook is not sufficient as shutdown of the process is
not inititated until all non-daemon threads have ended.

src/main/java/com/github/steveice10/packetlib/tcp/TcpClientSession.java
```diff
@@ -318,4 +318,15 @@ private static void createTcpEventLoopGroup() {
 
         Runtime.getRuntime().addShutdownHook(new Thread(() -> EVENT_LOOP_GROUP.shutdownGracefully()));
```
Camotoy: I think this was supposed to functionally do the same thing. Do you know why it didn't work?

Konicai (commented):
I feel like the method name could be more descriptive considering the EventLoopGroup is being shutdown. Also, should the three static fields relating to the EventLoopGroup be nullified in order to keep state consistent if someone tries to resume?

Also, it would probably be best to not catch the exception here?

src/main/java/com/github/steveice10/packetlib/tcp/TcpClientSession.java
petersv5: The shutdown hook is called when the process has decided to shut down. It will decide to do so when the last non-daemon thread ends. However, the thread pools will not end unless instructed, so this hook will be called when it is no longer needed.

petersv5:
I would suggest that someone that really understands the innards of Geyser rework this from scratch. This was very much a "poke this thing that seems broken" style of patch. 

The EventLoopGroup variables should be nullified if we (you) decide to go the route in this patch. Some of them were final and again - quick hack to verify that the worker pools were the culprits.

petersv5:
I guess if we null out the event loop group variables for the event loop groups that we have shut down then they will be recreated later if used again?

petersv5:
I pushed a new version that allows the TcpClientSessino to be restarted after a shutdown. This is an enabled for the corresponding update to Geyser to allow /geyser reload to work.

I also changed the name of the shutdown function to shutdownEventLoopGroup.

petersv5:
As mentioned in issue #699 this patch is only sufficient as long as the public static variable TcpSession.USE_EVENT_LOOP_FOR_PACKETS is set to false before instantiating the first TcpClientSession. This is the case with Geyser. Otherwise the inheritance from TcpSession will create another EventLoopGroup during object construction that is unused by TcpClientSession.

TL:DR the current PR is sufficient for Geyser, but the relationship with TcpSession should be considered for other used of ProtocolLib.

src/main/java/com/github/steveice10/packetlib/tcp/TcpClientSession.java
Camotoy: Then should we mark the threads a daemon? Can we?

petersv5: If I understand the discussion in the mailing lists for NEtty they switched to a daemon thread model in Netty rel 5.0. I am not sure what other implications there are of making them daemon threads. Possibly no negative effect at all. We would have to change the thread factory. Or switch to Netty 5, but I have no idea how big a change that is.

Another option is to move away from static variables / thread pools and instead have per-instance groups. Since we anyway don't have more than one user of TcpClientSession in Geyser no pooling gain would be lost as there simpliy is no pooling gain.

Camotoy: Moving to Netty 5 is not an option because Minecraft still uses Netty 4 and we depend on that in Geyser.
We have one TcpClientSession per GeyserSession class.

petersv5: I expected as much re netty 5.
Can we have more than one GeyserServer instance per process? If so, I guess it can be useful to have a global EventLoopGroup, but I'm not sure how useful it will be vs having one EventLoopGroup per TcpCLientSesstion instance which is shut down when the TcpClientSession instance is closed down.
It may be useful to test making a custom thread factory that creates daemon threads instead.

Camotoy: It seems like we either remove this line, or we figure out the easiest way to make the threads daemon.

petersv5: Is the shutdown hook ever needed? Do we need to shut down the EventLoopGroup as a part of the shutdown, or is it sufficient to just let them be killed as a part of the shutdown?

Camotoy: I specifically added the hook to try to mitigate this exact bug you're PRing to fix. So it sounds like this line is completely unnecessary.

petersv5: _If_ we decide to go with daemon threads then this line _may_ be needed to cleanly shut down the EventLoopGroup. I don't know. 

If we shut the EventLoopGroup down explicitly then this line is not needed.

I have not updated the patch pending directions from you about which way you want to take this.

Camotoy: I don't know what the side effects are, but it seems like setting the threads to daemon means we don't need to fret about tasks holding up the server, and we don't need "new API". I think our Bedrock Protocol already marks its threads as daemon.

petersv5: I can make a test version and see what happens...

petersv5: See alternative deamon thread based approach in #748.

Camotoy:
Closing in favor of #748.

PR #748, opened 2023-08-12
use daemon threads to allow shut down of application

petersv5:
This is an alternative to #744.

To exit normally all non-daemon threads of the java process are supposed to have stopped. TcpClientSession creates an static class member EventLoopGroup which in turn instantiates worker threads. These are changed to daemon threads to not get in the way of the shutdown. This is similar to the default in Netty 5.

To ensure work is finishd the Runtime shutdown event is hooked to initiate a time-limited graceful shutdown of the event loops. Normally an application should shut down communication before the process shutdown is inititated.


petersv5:
The TcpCLientSession change has been successfully tested with Geyser (not pushed to Geyser yet pending further discussion here). TcpSession is not tested yet.

petersv5:
This patch should resolve #699, needs testing.

Luluno01:
I'm not a heavy user of Java but just out of curious, how does providing a `DefaultThreadFactory` allow `shutdownGracefully` to shutdown all the threads while it didn't really work before? Does `DefaultThreadFactory` create daemon threads by default?

petersv5:
The second parameter to DefaultThreadFactory indicates daemon threads.

shutdownGracefully() has always worked. What did not work before was hooking the shutdown event to do the shutdownGracefully() as that event was never fired due to the lingering threads. Daemon threads do not veto shutdown, hence the Runtime object decides to shut down the process, calling the hooks etc.

Luluno01:
> 

Okay, thanks for the explanation! But would marking all worker threads as daemon threads hurt the performance when they do need to work, as daemon thread has lower priority to the best of my knowledge?

petersv5:
I don't think there is any change in priority when a thread is marked as daemon. I have not checked if  DefaultThreadFactory does that,  but it likely don't.  It is a pretty of netty, so knows it needs a reasonable priority.

Luluno01:
Oh yes I've verified their priorities by printing them out. They have the same priority as the main thread, which is 5 in my case. Sorry for the confusion.

petersv5:
Do we want to move forward with this direction and close #744?
Anything else that needs doing?

Luluno01:
Apparently they are too busy to review your code...

Konicai:
I wouldn't take it personally – the team does tend to be generally busy and it was the weekend. We really do appreciate what is being done here though <3

Some internal discussion was initiated but it should be wrapped up soon

petersv5:
I have no urgency, so no problem there for me. 
I was just curious which way we want to take this and if there was something else do be done.

Luluno01:
Not meant to blame anyone but was suggesting that it's probably going to be a while before the developers have time to handle this. Really appreciate such an open-source project where issues are actively discussed and solved.👍

petersv5:
Anything we can assist with here to help get it in shape before 1.20.2?

Konicai:
> The second parameter to DefaultThreadFactory indicates daemon threads.

> shutdownGracefully() has always worked. What did not work before was hooking the shutdown event to do the shutdownGracefully() as that event was never fired due to the lingering threads. Daemon threads do not veto shutdown, hence the Runtime object decides to shut down the process, calling the hooks etc.

Can we add a concise comment about the 2nd parameter and about the shutdown process to one of the classes? and then in the other class its probably just fine to refer to the other class for documentation about this.

petersv5:
> > The second parameter to DefaultThreadFactory indicates daemon threads.
> 
> > shutdownGracefully() has always worked. What did not work before was hooking the shutdown event to do the shutdownGracefully() as that event was never fired due to the lingering threads. Daemon threads do not veto shutdown, hence the Runtime object decides to shut down the process, calling the hooks etc.
> 
> Can we add a concise comment about the 2nd parameter and about the shutdown process to one of the classes? and then in the other class its probably just fine to refer to the other class for documentation about this.

Want me to add it, or will you guys take over updating the PR?
Either way is fine by me.

petersv5:
Pushed my proposal for comments.

Konicai (approved):

PR #750, opened 2023-08-15
Fix ClientboundLoginPacket serialization

WinPlay02:
The byte written after the gamemode should be the previous gamemode. The serialization now matches the deserialization.


basaigh (approved):
I can confirm this issue, as I had already fixed it in the 1.20.2 branch, but had forgotten to port it to master. This is good to merge.

PR #768, opened 2023-12-10
Label as 1.20.4 and update adventure to snapshot

Konicai:


onebeastchris (approved):
Quite unfortunate that adventure isn't out yet, but this looks good to me

Issue #676, opened 2021-12-04
DataPalette has an int field 'globalPaletteBits' which is not used


MatteCarra:
I was checking the ChunkSection.read method and I noticed it required a globalBiomePaletteBits.
I was not sure what it was and after inspecting the source I noticed that it was used to create a new DataPalette.
The problem is that it is not used there:
https://github.com/GeyserMC/MCProtocolLib/blob/acddfeeb51419abf06845a6fcd4fed14cacc4642/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/DataPalette.java#L21

Am I missing something?

Camotoy:
*Seems* like you're correct that it is unnecessary, yes.

onebeastchris:
Closing - this was resolved in the linked PR's.

PR #775, opened 2023-12-18
Fix DataPalette index & deprecate unused globalPaletteBits

### LINKED ISSUE #774: DataPalettes do not work correctly.
I'm getting an index out of bounds exception on coordinates where it should be fully writable to.

How to reproduce?

Create a new chunk section
`ChunkSection chunkSection = new ChunkSection();`

attempt to write to position (0, 1, 0) which should be valid since biome palettes should be 4x4x4
`chunkSection.getBiomeData().set(0, 1, 0, 5);`

And you will receive an index out of bounds exception.

davidmayr:
This PR attempts to fix #774 

From my basic testing, the block palettes still work as they did before without any issues and biome palettes do not throw any errors anymore. They also seem correct in game, but don't quote me on that. (Since my test world is very small with not that many different biomes (2 to be exact))

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/DataPalette.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/DataPalette.java
@@ -103,7 +103,7 @@ public class DataPalette {
         }
     }
 
-    private static int index(int x, int y, int z) {
-        return y << 8 | z << 4 | x;
+    private int index(int x, int y, int z) {
+        return y << paletteType.getMaxBitsPerEntry() | z << paletteType.getMinBitsPerEntry() | x;
     }
 }
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/DataPalette.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/DataPalette.java
@@ -66,8 +66,7 @@ public class DataPalette {
      */
     @Deprecated(forRemoval = true)
     public static DataPalette createEmpty(PaletteType paletteType, int globalPaletteBits) {
-        return new DataPalette(new ListPalette(paletteType.getMinBitsPerEntry()),
-                new BitStorage(paletteType.getMinBitsPerEntry(), paletteType.getStorageSize()), paletteType);
+        return createEmpty(paletteType);
     }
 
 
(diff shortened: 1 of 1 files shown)


AlexProgrammerDE:
I've found this in vanilla code:
image

davidmayr:
> I've found this in vanilla code:
> image
> 

That would result in a completly different index then before, right? I don't think that would work correctly unless other parts are rewritten aswell? Or am I wrong here?

davidmayr:
But I can try if it works in a few hours

Kas-tle:
These are equivalent as sizeBits is effectively maxBitsPerEntry minus minBitsPerEntry, so the position of the encoded bytes is the same.

For example, a chunk section is `CHUNK(4, 8, 4096)`, meaning a sizeBits of 4. Let's consider the position `[5, 7, 9]` in the chunk.

With the code in this PR, we would do:
```
( 7 << 8 ) | ( 9 << 4 ) | ( 5 )
11100000000b | 10010000b | 101b

11100000000
   10010000
        101
-----------
11110010101 = 1941
```

With the code from the Minecraft jar we would do:
```
( ( ( 7 << 4 ) | 9 ) << 4 ) | ( 5 )
( ( 1110000b | 1001b ) << 4 ) | ( 101b )

1110000
   1001
-------
1111001

( 1111001b << 4 ) | ( 101b )
11110010000b | 101b

11110010000
        101
-----------
11110010101 = 1941
```

Kas-tle:
Also since you're working in this area feel free to remove globalPaletteBits per https://github.com/GeyserMC/MCProtocolLib/issues/676

davidmayr:
> Also since you're working in this area feel free to remove globalPaletteBits per #676

Do you want me to make breaking API changes, or should I keep the existing static createEmpty, createForChunk, createForBiome methods and deprecate them?

Kas-tle:
Yes it should have a deprecation notice and for now just redirect to the new method

davidmayr:
> Yes it should have a deprecation notice and for now just redirect to the new method

And as for the @AllArgsConstructor ? Should I manually recreate a deprecated constructor with globalPaletteBits to not cause any breaking changes as well?

Kas-tle:
Correct. Basically the goal is just that we can easily remove it in the next breaking version.

davidmayr:
Done. Shouldn't involve any breaking changes now

Kas-tle (approved):

Kas-tle:
Thank you!

PR #780, opened 2023-12-23
Remove unused exceptions, use java 9+ features and checkerframework, other cleanup

AlexProgrammerDE:


src/main/java/com/github/steveice10/mc/protocol/codec/MinecraftCodecHelper.java
```diff
@@ -555,14 +549,10 @@ public void writeNumberFormat(ByteBuf buf, NumberFormat numberFormat) throws IOE
...
         PositionSourceType type = PositionSourceType.from(this.readVarInt(buf));
-        switch (type) {
-            case BLOCK:
-                return new BlockPositionSource(this.readPosition(buf));
-            case ENTITY:
-                return new EntityPositionSource(this.readVarInt(buf), buf.readFloat());
-            default:
-                throw new IllegalStateException("Unknown position source type!");
-        }
+        return switch (type) {
```
onebeastchris, line 550: why was the default case removed here?

src/main/java/com/github/steveice10/mc/protocol/data/game/entity/metadata/ByteMetadataType.java
```diff
@@ -10,7 +10,7 @@ public class ByteMetadataType extends MetadataType<Byte> {
     private final ByteReader primitiveReader;
     private final ByteWriter primitiveWriter;
     private final ByteEntityMetadataFactory primitiveFactory;
-    
+
```
onebeastchris, line 11: ?

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/ClientboundAwardStatsPacket.java
```diff
@@ -27,44 +23,23 @@ public ClientboundAwardStatsPacket(ByteBuf in, MinecraftCodecHelper helper) {
...
-                case KILLED_BY_ENTITY:
-                    statistic = new KilledByEntityStatistic(EntityType.from(statisticId));
-                    break;
-                case CUSTOM:
-                    statistic = CustomStatistic.from(statisticId);
-                    break;
-                default:
-                    throw new IllegalStateException();
-            }
+            Statistic statistic = switch (category) {
```
onebeastchris, line 26: let's keep the default case too?

onebeastchris (changes requested):
Looks good so far, just a few tiny things

src/main/java/com/github/steveice10/mc/protocol/data/game/entity/metadata/ByteMetadataType.java
AlexProgrammerDE: Waste of bytes and file size.

src/main/java/com/github/steveice10/mc/protocol/codec/MinecraftCodecHelper.java
AlexProgrammerDE: This code will not compile if not all enum values are implemented.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/ClientboundAwardStatsPacket.java
AlexProgrammerDE: This code will not compile if not all enum values are implemented.

onebeastchris: Sure, but only if all enums are implemented... I'd rather have a default throw that's not needed most of the time that can catch issues/missed updates/unsupported modded scenarios 
It's what we do with Geyser atleast, I don't see why it wouldn't fit here

AlexProgrammerDE: Enums are not to be extended at runtime.

onebeastchris: That's not what I'm concerned about, I'm aware of that :)
A default case could e.g. warn about a missed new enum that may not be implemented in MCPL when there's a Minecraft update.

AlexProgrammerDE: It doesn't, it just throws a exception at runtime. In this case the code does not compile when you missed handling a new enum option.

onebeastchris: that still assumes the enum was added to mcpl :p
As long as some exception/warning is thrown at runtime when something isn't known, I'm fine with it

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/serverbound/inventory/ServerboundSetCreativeModeSlotPacket.java
```diff
@@ -16,7 +16,7 @@
 @AllArgsConstructor
 public class ServerboundSetCreativeModeSlotPacket implements MinecraftPacket {
     private final int slot;
-    private final @NonNull ItemStack clickedItem;
+    private final ItemStack clickedItem;
```
onebeastchris: If this is nullable, then please add the annotation

.editorconfig
Kas-tle, line 1: Please gitignore this file

onebeastchris (commented):
lgtm, just the one annotation

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/serverbound/inventory/ServerboundSetCreativeModeSlotPacket.java
AlexProgrammerDE: Will do.

.editorconfig
AlexProgrammerDE: Actually, I recommend committing this file. It helps contributors to use the recommended code style. It's a standard file that most editors accept and use.

AlexProgrammerDE: Here is a example of a project using it: https://github.com/PaperMC/paperweight-test-plugin/blob/master/.editorconfig

AlexProgrammerDE: And here: https://github.com/PaperMC/Paper/blob/master/.editorconfig

Tim203: While I'm not against adding this, it should be noted that the `ij_` prefixed option(s) are IntelliJ/JetBrains specific. So while I like the idea of the project following some styling guidelines, I personally think that editorconfig is too weak to be useful. Which is why I personally stopped adding it to projects.

AlexProgrammerDE: Yes that is there to have it be the way konicai wants it for all code in the codebase and in PRs: https://github.com/GeyserMC/MCProtocolLib/pull/769#discussion_r1428561086

onebeastchris: That comment just mentions to avoid using wildcard imports - which you can disable in your intelliJ settings - but it doesn't mention adding an editorconfig, which currently isn't something that's been agreed upon internally yet.

AlexProgrammerDE: Actually the way konicai recommends it was setting the setting to 1000, which this editorconfig does. Konicai didn't tell me to add a editorconfig, but I added it myself, so other contributors also have this code style setting in intellij. That's the meaning of the line `ij_java_class_count_to_use_import_on_demand = 1000`
Everything else in there is intellij defaults anyways, so it changes pretty much nothing except that one thing konicai wants all contributors to do.

onebeastchris: The point is that this editorconfig only does that for IntelliJ, and not other ide's. Hence why we'd use a different solution - once we decide on common styling guidelines - to enforce them that's more universal. You could gitignore the file for now. Imo it's not a good idea to have the file for the sake of one editor specific setting

onebeastchris (approved):

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/palette/MapPalette.java
```diff
@@ -17,19 +16,21 @@
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 @EqualsAndHashCode
 public class MapPalette implements Palette {
-    private final int maxId;
+    private static final int MISSING_ID = -1;
+    private final int size;
```
Konicai: can we do `capacity` instead

Konicai (changes requested):

src/main/java/com/github/steveice10/mc/protocol/data/game/chunk/palette/MapPalette.java
AlexProgrammerDE: It's the mojang name last time I checked. I'll verify again later.

Konicai (approved):

PR #781, opened 2023-12-28
Fix ClientboundSetTitleTextPacket serialization

WinPlay02:
The title component should being encoded as an NBT tag instead of a string.

This should fix the 1.20.4 vanilla client disconnecting when connecting with an exception.
The title component is now NotNull.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/title/ClientboundSetTitleTextPacket.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/title/ClientboundSetTitleTextPacket.java
@@ -15,7 +15,7 @@ import java.io.IOException;
 @With
 @AllArgsConstructor
 public class ClientboundSetTitleTextPacket implements MinecraftPacket {
-    private @NotNull final Component text;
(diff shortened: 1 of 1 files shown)


Konicai:
thank you

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/title/ClientboundSetTitleTextPacket.java
```diff
@@ -2,28 +2,27 @@
...
+import org.jetbrains.annotations.NotNull;
 
 import java.io.IOException;
 
 @Data
 @With
 @AllArgsConstructor
 public class ClientboundSetTitleTextPacket implements MinecraftPacket {
-    private final @Nullable Component text;
+    private final @NotNull Component text;
```
Konicai: Sorry, you can use the lombok annotation which should insert a null check to the constructor

WinPlay02: Done

Konicai (approved):

PR #784, opened 2024-01-08
Add coverage

AlexProgrammerDE:


Kas-tle:
Is this still needed, and if so for what purpose?

AlexProgrammerDE:
It's not "needed". It's a nice-to-have. Coverage checks how much of the library code you're actually testing inside of your tests. So like a coverage of 25% means you test 25% of your library code inside unit tests.

Kas-tle (approved):

PR #787, opened 2024-01-17
Throw when trying to write an unknown block entity type

onebeastchris:
This marks the type fields in BlockEntityInfo and ClientboundBlockEntityDataPacket as being nullable, and throws when trying to write a null type.

### DIFF
--- a/src/main/java/com/github/steveice10/mc/protocol/codec/MinecraftCodecHelper.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/codec/MinecraftCodecHelper.java
@@ -626,6 +626,9 @@ public class MinecraftCodecHelper extends BasePacketCodecHelper {
     }
 
     public void writeBlockEntityType(ByteBuf buf, BlockEntityType type) {
+        if (type == null) {
+            throw new IllegalArgumentException("Cannot write an unknown BlockEntityType!");
+        }
         this.writeEnum(buf, type);
     }
 
--- a/src/main/java/com/github/steveice10/mc/protocol/data/game/level/block/BlockEntityInfo.java
+++ b/src/main/java/com/github/steveice10/mc/protocol/data/game/level/block/BlockEntityInfo.java
@@ -11,6 +11,6 @@ public class BlockEntityInfo {
     private final int x;
     private final int y;
     private final int z;
-    private final BlockEntityType type;
+    private final @Nullable BlockEntityType type;
     private final @Nullable CompoundTag nbt;
 }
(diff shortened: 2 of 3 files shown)


src/main/java/com/github/steveice10/mc/protocol/codec/MinecraftCodecHelper.java
```diff
@@ -626,6 +626,9 @@ public BlockEntityType readBlockEntityType(ByteBuf buf) {
     }
 
     public void writeBlockEntityType(ByteBuf buf, BlockEntityType type) {
+        if (type == null) {
+            throw new IllegalArgumentException("Cannot write an unknown BlockEntityType!");
+        }
```
Konicai, line 631: Annotate the `type` parameter accordingly

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/level/ClientboundBlockEntityDataPacket.java
```diff
@@ -19,7 +19,7 @@
 @AllArgsConstructor
 public class ClientboundBlockEntityDataPacket implements MinecraftPacket {
     private final @NonNull Vector3i position;
-    private final @NonNull BlockEntityType type;
+    private final @Nullable BlockEntityType type;
```
Konicai, line 22: With lombok, the constructor parameter will also be annotated as Nullable, even though an IllegalArgumentException will be thrown when this packet is serialized.

src/main/java/com/github/steveice10/mc/protocol/data/game/level/block/BlockEntityInfo.java
```diff
@@ -11,6 +11,6 @@ public class BlockEntityInfo {
     private final int x;
     private final int y;
     private final int z;
-    private final BlockEntityType type;
+    private final @Nullable BlockEntityType type;
```
Konicai, line 14: Null is illegal here: https://github.com/GeyserMC/MCProtocolLib/blob/master/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/level/ClientboundLevelChunkWithLightPacket.java#L60

Konicai (changes requested):
We might want to consider allowing writing null BlockEntityTypes.

Otherwise, we  *could* have the constructor parameters be annotated differently. But a proxy using MCPL wouldn't be able to deserialize and re-serialize the packets with BlockEntityType, which doesn't seem ideal.

src/main/java/com/github/steveice10/mc/protocol/packet/ingame/clientbound/level/ClientboundBlockEntityDataPacket.java
onebeastchris: What would you suggest as an alternative? No NonNull/Nullable marking?

PR #792, opened 2024-02-28
Velocity natives

AlexProgrammerDE:


Camotoy:
Velocity is GPL 3.0 meaning that we *technically* cannot use its natives without switching MCProtocolLib to GPL. Note that https://github.com/astei/libdeflate-java I think is similar, and is licensed as MIT.

AlexProgrammerDE:
Won't work due to Velocity being GPL

PR #798, opened 2024-04-24
1.20.5 dev 2

AlexProgrammerDE:


protocol/src/main/java/org/geysermc/mcprotocollib/protocol/codec/MinecraftCodecHelper.java
```diff
@@ -456,82 +586,75 @@ public void writeParticle(ByteBuf buf, Particle particle) {
...
-            case SHRIEK -> {
-                ShriekParticleData shriek = (ShriekParticleData) data;
-                this.writeVarInt(buf, shriek.getDelay());
-            }
-            case VIBRATION -> {
-                VibrationParticleData vibration = (VibrationParticleData) data;
-                this.writePositionSource(buf, vibration.getPositionSource());
-                this.writeVarInt(buf, vibration.getArrivalTicks());
-            }
+            case BLOCK:
```
Camotoy: This seems like a regression, since we're moving from new-style switch blocks to old-style switch blocks.

Camotoy (approved):

PR #808, opened 2024-05-08
Add Slf4j

AlexProgrammerDE:


Camotoy:
I'd rather not use Lombok, and would prefer for the logger to just be created statically, unless there's an easy benefit to Lombok log creation. Mainly it should help any case clear where the logger is no longer used by marking the variable unused. Plus, it's one line, if I'm correct.

Camotoy:
Once we finalize using Lombok or not here, I'll merge.

AlexProgrammerDE:
Removed the annotation

PR #809, opened 2024-05-09
Added color In BossBarColor class

DionCoder:
Added:

Color **BLUE** in BossBarColor class


Camotoy:
There already is a blue boss bar color - it's likely the "cyan" already present.

Camotoy:
Thank you anyway for the PR! If you think the enum should be renamed, go ahead and make an issue for it.

PR #811, opened 2024-05-14
Use built-in ResourceLocation type

AlexProgrammerDE:


Camotoy:
I'm neutral on this and will delegate it to Bas or someone else to merge. I absolutely know that Keys are easier to use instead of loads of string splitting, but I did some testing and they do take up more memory, even with the `minecraft` prefix shared across all Key instances. So it's a tradeoff, and it's not one I'd prefer to push the merge button on.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/data/game/entity/attribute/AttributeType.java
```diff
@@ -18,44 +19,44 @@ public interface AttributeType {
...
+        GENERIC_JUMP_STRENGTH(Key.key("minecraft:generic.jump_strength"), 0.42, 0, 32),
+        GENERIC_KNOCKBACK_RESISTANCE(Key.key("minecraft:generic.knockback_resistance"), 0, 0, 1),
+        GENERIC_LUCK(Key.key("minecraft:generic.luck"), 0, -1024, 1024),
+        GENERIC_MAX_ABSORPTION(Key.key("minecraft:generic.max_absorption"), 0, 0, 2048),
+        GENERIC_MAX_HEALTH(Key.key("minecraft:generic.max_health"), 20, 1, 1024),
+        GENERIC_MOVEMENT_SPEED(Key.key("minecraft:generic.movement_speed"), 0.7F, 0, 1024),
+        GENERIC_SAFE_FALL_DISTANCE(Key.key("minecraft:generic.safe_fall_distance"), 3, -1024, 1024),
+        GENERIC_SCALE(Key.key("minecraft:generic.scale"), 1, 0.0625, 16),
+        ZOMBIE_SPAWN_REINFORCEMENTS(Key.key("minecraft:zombie.spawn_reinforcements"), 0, 0, 1),
+        GENERIC_STEP_HEIGHT(Key.key("minecraft:generic.step_height"), 0.6, 0, 10);
```
basaigh: Probably easier to add a constructor which handles the `Key.key()` rather than calling it so many times.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/codec/MinecraftCodecHelper.java
```diff
@@ -161,12 +161,13 @@ public <T> void writeHolder(ByteBuf buf, Holder<T> holder, BiConsumer<ByteBuf, T
...
-        return Identifier.formalize(this.readString(buf));
+    @SuppressWarnings("PatternValidation")
+    public Key readResourceLocation(ByteBuf buf) {
+        return Key.key(this.readString(buf));
     }
 
-    public void writeResourceLocation(ByteBuf buf, String location) {
-        this.writeString(buf, location);
+    public void writeResourceLocation(ByteBuf buf, Key location) {
+        this.writeString(buf, location.toString());
```
basaigh: Replace with `.asString()`?

basaigh (changes requested):

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/data/game/entity/attribute/AttributeType.java
AlexProgrammerDE: Hm sure, but then I'd need to add @KeyPattern to the constructor arg to ensure they actually have the right name. Is that fine?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/codec/MinecraftCodecHelper.java
AlexProgrammerDE: It just delegates to it afaik, but sure

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/data/game/entity/attribute/AttributeType.java
basaigh: Do you? It's not like it would ever be called outside of the class.

AlexProgrammerDE: Yeah or else I'd need to add a @SuppressWarnings("PatternValidation")

basaigh: In which case, yeah, fair enough.

AlexProgrammerDE: Alright

basaigh (approved):

PR #814, opened 2024-05-16
Fix writing mobeffectinstances

AlexProgrammerDE:


AlexProgrammerDE:
mobeffectdetails don't actually store the effect type. Nested details thus don't contain the effect multiple times. Only the MobEffectInstance has the effect type.

AJ-Ferguson (approved):

PR #816, opened 2024-05-17
Better modifieroperation

AlexProgrammerDE:


protocol/src/main/java/org/geysermc/mcprotocollib/protocol/data/game/item/component/ItemAttributeModifiers.java
```diff
@@ -32,8 +32,8 @@ public static class AttributeModifier {
 
     public enum EquipmentSlotGroup {
         ANY,
-        MAIN_HAND,
-        OFF_HAND,
+        MAINHAND,
+        OFFHAND,
```
Redned235, line 36: I think the former names were better here. I don't really see the benefit of following vanilla names perfectly for the sake of doing so (especially, when the difference is an underscore).

Redned235 (changes requested):

PR #820, opened 2024-05-22
Allow using arbitrary json as the status packet

AlexProgrammerDE:
I'd prefer using any arbitrary jsonobject as the status packet payload to support third-party modloader fields.
The implementation keeps support for the typed json data, but allows the user to additionally provide any json object they want.
I don't think this is outside of the scope of the project as this could allow people building a MCPL client (like me) to detect if a server to connect to is a forge server to enable different logic/prevent joining due to different protocol. 
my current workaround is registering my own status data packet, which is a thing i wanna stop doing

### DIFF
--- a/protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/status/clientbound/ClientboundStatusResponsePacket.java
+++ b/protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/status/clientbound/ClientboundStatusResponsePacket.java
@@ -31,13 +31,25 @@ public class ClientboundStatusResponsePacket implements MinecraftPacket {
     // vanilla behavior falls back to false if the field was not sent
     private static final boolean ENFORCES_SECURE_CHAT_DEFAULT = false;
 
-    private final @NonNull ServerStatusInfo info;
+    private final @NonNull JsonObject data;
+
+    public ClientboundStatusResponsePacket(ServerStatusInfo info) {
+        this(toJson(info));
+    }
 
     public ClientboundStatusResponsePacket(ByteBuf in, MinecraftCodecHelper helper) {
-        JsonObject obj = new Gson().fromJson(helper.readString(in), JsonObject.class);
-        JsonElement desc = obj.get("description");
+        data = new Gson().fromJson(helper.readString(in), JsonObject.class);
+    }
+
+    @Override
+    public void serialize(ByteBuf out, MinecraftCodecHelper helper) {
+        helper.writeString(out, data.toString());
+    }
+
+    public ServerStatusInfo getInfo() {
+        JsonElement desc = data.get("description");
         Component description = DefaultComponentSerializer.get().serializer().fromJson(desc, Component.class);
-        JsonObject plrs = obj.get("players").getAsJsonObject();
+        JsonObject plrs = data.get("players").getAsJsonObject();
         List<GameProfile> profiles = new ArrayList<>();
         if (plrs.has("sample")) {
             JsonArray prof = plrs.get("sample").getAsJsonArray();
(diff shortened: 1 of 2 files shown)


protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/status/clientbound/ClientboundStatusResponsePacket.java
```diff
@@ -31,13 +31,25 @@ public class ClientboundStatusResponsePacket implements MinecraftPacket {
     // vanilla behavior falls back to false if the field was not sent
     private static final boolean ENFORCES_SECURE_CHAT_DEFAULT = false;
 
-    private final @NonNull ServerStatusInfo info;
+    private final @NonNull JsonObject data;
+
+    public ClientboundStatusResponsePacket(ServerStatusInfo info) {
+        this(toJson(info));
+    }
```
Konicai, line 38: `@NonNull` on the param here and on the wither too

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/status/clientbound/ClientboundStatusResponsePacket.java
```diff
@@ -31,13 +31,25 @@ public class ClientboundStatusResponsePacket implements MinecraftPacket {
...
-        JsonElement desc = obj.get("description");
+        data = new Gson().fromJson(helper.readString(in), JsonObject.class);
+    }
+
+    @Override
+    public void serialize(ByteBuf out, MinecraftCodecHelper helper) {
+        helper.writeString(out, data.toString());
+    }
+
+    public ServerStatusInfo getInfo() {
```
Konicai: I understand this was named as so to avoid breaking changes, but I'm still gonna propose naming it something like `parseInfo` to indicate that it shouldn't be called repeatedly

Konicai (changes requested):
looks good

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/status/clientbound/ClientboundStatusResponsePacket.java
```diff
@@ -31,13 +31,25 @@ public class ClientboundStatusResponsePacket implements MinecraftPacket {
...
+    private final @NonNull JsonObject data;
+
+    public ClientboundStatusResponsePacket(@NonNull ServerStatusInfo info) {
+        this(toJson(info));
+    }
 
     public ClientboundStatusResponsePacket(ByteBuf in, MinecraftCodecHelper helper) {
-        JsonObject obj = new Gson().fromJson(helper.readString(in), JsonObject.class);
-        JsonElement desc = obj.get("description");
+        data = new Gson().fromJson(helper.readString(in), JsonObject.class);
```
Camotoy, line 41: This can probably be `new JsonParser().parse` - it may say that's deprecated but I know Geyser would much appreciate using the deprecated version instead of the modern version not available on older Gson versions.

AlexProgrammerDE: Isn't Gson.fromJson available in old versions?

AlexProgrammerDE: @Camotoy ?

Camotoy: Yes, but JsonParser (available since 1.3) is probably a *tad* bit cleaner and Gson doesn't have to look up the type for JsonObject like with fromJson. I see the old code used fromJson as well, so it doesn't really matter.

AlexProgrammerDE: I see, but this ensures upwards support. Newer Gson versions may remove that deprecated method and that's why I wanna avoid using it.

Camotoy: It's been deprecated for five years and there's little reason for Gson to remove it. In a couple years when no project will be using an outdated Gson, we can switch to the static version and it won't be an issue.

AlexProgrammerDE: Tried adding it now, but I don't think it'll work without doing workarounds to use Jsonparser.

AlexProgrammerDE: The main issue is that JsonParser#parse actually generates JsonElements, not JsonObjects. So I'd have to make casts. I think Gson can provide nicer error messages than a cast error if we use fromJson.

PR #821, opened 2024-05-23
Fix optional status fields

AlexProgrammerDE:
Builds upon #820 
Fixes reading server stats info from vanilla server that have either version or players missing.
A few small cleanups were also made. Every change has its own commit.


onebeastchris (approved):
lgtm

PR #825, opened 2024-06-13
Switch to Mojang mapped level event names

basaigh:


onebeastchris (approved):
yes please - we don' want to end up on https://mappings.cephx.dev/ :p

PR #826, opened 2024-06-16
Implementation for inbound and outbound splitting

AlexProgrammerDE:
This PR basically splits PacketProtocol into PacketProtocol and packetRegistry. A PacketProtocol defines basic stuff about the protocol like SRV and it provides incoming and outgoing Packet registries. A PacketRegistry only contains packet id mappings and everything related to the actual packet classes. The MinecraftProtocol extends the PacketProtocol and allows dynamically switching the PacketRegistries that are incoming and outgoing.


AlexProgrammerDE:
I've looked deeper and it seems packets all initially get handled on the netty thread and the *conditionally* get moved away onto the main tick thread. This does of course create issues with the way MCPLs impl works and may lead to different behaviour compared to vanilla on a few packets since they were expected to run on the netty thread, but actually got offloaded to the packet eventloop.

AlexProgrammerDE:
This PR is read-only and not supposed to be merged. While it is fixing packet race conditions, it will be replaced by many smaller PRs focused on aspects of splitting. This PR will be closed once different PRs have added all its functionality. This PR is only a reference for myself on a working impl.

AlexProgrammerDE:
Not needed anymore, the new impl is coded.

PR #827, opened 2024-06-16
Translatable messages

AlexProgrammerDE:
Based upon #826


AlexProgrammerDE:
I'm moving this into another PR detached from protocol switching. That way this can be merged quickly.

PR #828, opened 2024-06-16
Optimize stack

AlexProgrammerDE:


Kas-tle:
Can you explain why these changes are needed?

AlexProgrammerDE:
As explained on Discord, the main benefits are reduced allocations, using newer and faster Java 11+ NIO APIs and improved error-handling.

AlexProgrammerDE:
I'm working on a little extra stuff for this PR. We can heavily simplify the entire pipeline by switching to AttributeKeys for controlling encryption/compression from netty. That way we don't need to modify the pipeline after initialization.

AlexProgrammerDE:
Nah, lets get merged first what is already here. I don't want the PR to get too big.But I roughly know how this should be implemented. Maybe I'll make it into another PR.

protocol/src/main/java/org/geysermc/mcprotocollib/network/compression/ZLIBCompression.java
```diff
@@ -0,0 +1,77 @@
+package org.geysermc.mcprotocollib.network.compression;
+
+import io.netty.buffer.ByteBuf;
+
+import java.nio.ByteBuffer;
+import java.util.zip.DataFormatException;
+import java.util.zip.Deflater;
+import java.util.zip.Inflater;
+
+public class ZLIBCompression implements PacketCompression {
```
Redned235:
```suggestion
public class ZlibCompression implements PacketCompression {
```

Redned235 (changes requested):

Tim203:
Do you have some benchmarks of the before and after?

AlexProgrammerDE:
> Do you have some benchmarks of the before and after?

Unfortunately I don't have the tooling right now to test this properly. 
But I can tell you about why this is faster. 
For compression/decompression now nio buffers are used. That reduces the array allocations and works more closely with the deflater/inflater classes. That way operations are slightly faster.
For encryption the same methods are used, but it's faster because of removal of unnecessary array size allocations.
With the encryption algorithm used in mc the in and out size of the algorithm is the same and thus no extra arrays need to be allocated, we can simply just reuse the input buffer.

protocol/src/main/java/org/geysermc/mcprotocollib/network/crypt/PacketEncryption.java
```diff
@@ -2,24 +2,9 @@
 
 /**
  * An interface for encrypting packets.
+ * The outputLength should always be the same as the inputLength.
```
onebeastchris, line 5: Can we add a comment here, that this is vanilla behavior?

AlexProgrammerDE: That's the behaviour that the algorithm vanilla uses has. Ig I can add a comment saying that this is a trait of the algo vanilla uses.

protocol/src/main/java/org/geysermc/mcprotocollib/network/compression/ZlibCompression.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public ZlibCompression(int level) {
+        this.deflater = new Deflater(level);
+        this.inflater = new Inflater();
+    }
+
+    @Override
+    public void inflate(ByteBuf source, ByteBuf destination, int uncompressedSize) throws DataFormatException {
+        final int origIdx = source.readerIndex();
```
Konicai: `originalIndex`, similarly with deflate

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCompression.java
```diff
@@ -1,88 +1,73 @@
...
-            this.deflater.setInput(bytes, 0, readable);
-            this.deflater.finish();
-            while (!this.deflater.finished()) {
-                int length = this.deflater.deflate(this.buf);
-                out.writeBytes(this.buf, 0, length);
-            }
-
-            this.deflater.reset();
+            this.session.getCodecHelper().writeVarInt(outBuf, uncompressed);
+            compression.deflate(msg, outBuf);
```
Konicai: we don't want to check for the max uncompressed size anymore?

Konicai (changes requested):

protocol/src/main/java/org/geysermc/mcprotocollib/network/crypt/PacketEncryption.java
```diff
@@ -28,10 +13,9 @@ public interface PacketEncryption {
...
      * @param output Array to output decrypted data to.
      * @param outputOffset Offset of the output array to start at.
-     * @return The number of bytes stored in the output array.
      * @throws Exception If an error occurs.
      */
-    int decrypt(byte[] input, int inputOffset, int inputLength, byte[] output, int outputOffset) throws Exception;
+    void decrypt(byte[] input, int inputOffset, int inputLength, byte[] output, int outputOffset) throws Exception;
 
     /**
      * Encrypts the given data.
```
Konicai, line 23: mention that the input and output arrays may be the same instance, in which case the read and write regions may overlap

question: with the PacketEncryptor no longer taking the output length into account, is there a chance that the input array could ever be too small for the output array? [Cipher#update](https://docs.oracle.com/javase/8/docs/api/javax/crypto/Cipher.html#update-byte:A-int-int-byte:A-int-)

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCompression.java
AlexProgrammerDE: We do using         if (uncompressed < this.session.getCompressionThreshold()) { 
Or is there smth I'm missing?

AlexProgrammerDE: Oh I see now.

protocol/src/main/java/org/geysermc/mcprotocollib/network/crypt/PacketEncryption.java
AlexProgrammerDE: No, with AES/CFB8/NoPadding input size == output size

onebeastchris (approved):
lgtm

Redned235 (approved):

Konicai (approved):

PR #831, opened 2024-06-18
Use translatable components like vanilla and cleanup type casts

AlexProgrammerDE:
Just a few messages becoming dynamic using translatable components like vanilla instead of being hardcoded.


AlexProgrammerDE:
This also removes the option not to specify a reason component or specifying a String as a reason. The reason for this change is to enforce providing a clear reason to the client. Developers who want to provide a String reason need to wrap it using Component.text(). Developers who want to provide no disconnect reason can just provide an empty reason like Component.text("Connection closed."). This should really not be done and a proper reason component should be provided when working on a client or server.

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java
```diff
@@ -279,22 +277,12 @@ public void send(Packet packet) {
...
-
-    @Override
-    public void disconnect(Component reason) {
+    public void disconnect(@NonNull Component reason) {
         this.disconnect(reason, null);
     }
 
     @Override
-    public void disconnect(final Component reason, final Throwable cause) {
+    public void disconnect(@NonNull Component reason, @Nullable Throwable cause) {
```
onebeastchris, line 280: We use checkerframework's NonNull/Nullable annotations in other projects as a standard; let's also do so here?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -125,8 +125,7 @@ public void packetReceived(Session session, Packet packet) {
                 PrivateKey privateKey = KEY_PAIR.getPrivate();
 
                 if (!Arrays.equals(this.challenge, keyPacket.getEncryptedChallenge(privateKey))) {
-                    session.disconnect("Invalid challenge!");
-                    return;
+                    throw new IllegalStateException("Protocol error");
```
onebeastchris, line 128: Why are we not disconnecting the session here first?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -183,6 +182,7 @@ public void packetReceived(Session session, Packet packet) {
                 protocol.setState(ProtocolState.CONFIGURATION);
             } else if (packet instanceof ServerboundPingRequestPacket pingRequestPacket) {
                 session.send(new ClientboundPongResponsePacket(pingRequestPacket.getPingTime()));
+                session.disconnect(Component.translatable("multiplayer.status.request_handled"));
```
onebeastchris, line 185: What's the reasoning here?

onebeastchris (changes requested):

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: It's what vanilla does. It's seen as an unexpected error because the client sent an invalid payload.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: image
That's just what vanilla does

AlexProgrammerDE: nvm wrong direction, lemme get you the right one

AlexProgrammerDE: image
image

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java
AlexProgrammerDE: But packets have lomboks NonNull. But sure I'll do it.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
onebeastchris: fair enough

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
onebeastchris: Will the thrown exception result in a session disconnect? If not, i'd definitely do that first before throwing/returning
IMO we don't need to follow vanilla exactly, especially for cases like these

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java
onebeastchris: We'll probably want to change that as well in the future :p

AlexProgrammerDE: The nice thing about lomboks annotations is that they generate enforced null checks at compile time.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: Depends whether it runs on the netty thread or the event loop. Worst case, it just times out and prints an error in the console.

AlexProgrammerDE: It's basically the same level as us throwing decryption errors, which the key packet already does.

onebeastchris: A timeout seems sub-optimal; i'd rather keep the current disconnect behavior to ensure that cannot happen

AlexProgrammerDE: Like for example `getEncryptedChallenge` could error with an `IllegalStateException` (Explicitly thrown too) because the bytes are incorrect. Same goes for this case.

AlexProgrammerDE: Actually, NVM. It doesn't depend on the loop. It'll cause an immediate disconnect with an "Internal Exception" message with an error message. 
image

onebeastchris: Okay, that sounds fair too; but let's hear what others think

Konicai: Seems alright - looks like any other SessionListeners will be skipped which is good. Keeping the current message seems more developer friendly though

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java
```diff
@@ -385,21 +371,17 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
-        String message;
-        if (cause instanceof ConnectTimeoutException || (cause instanceof ConnectException && cause.getMessage().contains("connection timed out"))) {
-            message = "Connection timed out.";
-        } else if (cause instanceof ReadTimeoutException) {
-            message = "Read timed out.";
-        } else if (cause instanceof WriteTimeoutException) {
-            message = "Write timed out.";
+        Component message;
```
Konicai, line 375: Is it actually worth it to remove the granularity here?

AlexProgrammerDE: It's translated and the mc server doesn't distinguish them. I think it's worth it. There's no benefit to the end user whether it's a read or write timeout. On write timeout the client would disconnect with a client-side read-timeout anyways.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: yup, I don't think branching off vanilla behavior benefits anyone here. It makes comparing the code directly with vanilla harder for an edge case that'll basically be never be called.

onebeastchris (approved):

Konicai (approved):
thanks, pretty nice to see this cleaned up

PR #832, opened 2024-06-18
Streamline api

AlexProgrammerDE:
This basically makes MinecraftCodec stateless and removes unused APIs.
The MinecraftCodecHelper has a state for an unused registry API, which is not needed at all. Developers can just send their own registry as seen in ServerListener. These only exist because there was an attempt at allowing multi-version support inside MinecraftCodecHelper, but that's not gonna be necessary if ViaVersion support gets added to MCPL.
The option to add back logic for multi-version support is fully open for the future. I didn't remove MinecraftCodecHelper being instance-based, so it's fully possible to readd the multi-version logic back in the future. Just level events and sounds being dynamic is not enough for multi-version support and thus should be removed to streamline the API.
This also removes many workarounds from the code that cause possible inconsistency. Often the MinecraftCodecHelper gets initialized in the tests with empty parameters. This removes them as parameters and removes the need to provide them by moving them to the object itself.
PacketCodecs were also improved to decrease the inheritance issues and streamline the code flow and remove the UnsupportedOperationExceptions being needed.
BufferedPacket was also removed since it's useless for Minecraft.


Camotoy (approved):

basaigh (approved):
There do seem to be some unrelated changes present, but since this is seemingly required for state switching to be fixed, approved.

Issue #837, opened 2024-07-05
Invalid session on valid session.


tbvns:
When I generate a session token with 
```java
new StepMsaDeviceCode.MsaDeviceCodeCallback(msaDeviceCode -> {
       ChatManager.managers.get(proxiedSession).addEphemeral("Go to " + msaDeviceCode.getDirectVerificationUri(), 70);
       System.out.println(msaDeviceCode.getDirectVerificationUri());
}));
```
I get a valid session token (I can log in with a client that support token-logging) but I still can't log in to a premium server like Hypixel with my bot. I still get an invalid session when the session is clearly valid.
Note that I also tried with the code in [MinecraftProtocolTest.java](https://github.com/GeyserMC/MCProtocolLib/blob/master/example/src/main/java/org/geysermc/mcprotocollib/protocol/example/MinecraftProtocolTest.java), and I have the same issue.

tbvns:
Here is my code:
```java
package xyz.tbvns;

import net.kyori.adventure.text.Component;
import net.raphimc.minecraftauth.MinecraftAuth;
import net.raphimc.minecraftauth.step.java.StepMCProfile;
import net.raphimc.minecraftauth.step.java.StepMCToken;
import net.raphimc.minecraftauth.step.java.session.StepFullJavaSession;
import net.raphimc.minecraftauth.step.msa.MsaCodeStep;
import net.raphimc.minecraftauth.step.msa.StepCredentialsMsaCode;
import net.raphimc.minecraftauth.step.msa.StepMsaDeviceCode;
import net.raphimc.minecraftauth.util.OAuthEnvironment;
import org.geysermc.mcprotocollib.auth.GameProfile;
import org.geysermc.mcprotocollib.auth.SessionService;
import org.geysermc.mcprotocollib.network.ProxyInfo;
import org.geysermc.mcprotocollib.network.Session;
import org.geysermc.mcprotocollib.network.event.session.*;
import org.geysermc.mcprotocollib.network.packet.DefaultPacketHeader;
import org.geysermc.mcprotocollib.network.packet.Packet;
import org.geysermc.mcprotocollib.network.tcp.TcpClientSession;
import org.geysermc.mcprotocollib.protocol.MinecraftConstants;
import org.geysermc.mcprotocollib.protocol.MinecraftProtocol;
import org.geysermc.mcprotocollib.protocol.codec.MinecraftCodec;
[...]

PR #838, opened 2024-07-18
Fix edge-case where Texture#toString has a NullPointer because gson s…

AlexProgrammerDE:
…et a null metadata

Also changes NORMAL name to WIDE since that's the accurate name used inside the mc client.

### DIFF
--- a/protocol/src/main/java/org/geysermc/mcprotocollib/auth/GameProfile.java
+++ b/protocol/src/main/java/org/geysermc/mcprotocollib/auth/GameProfile.java
@@ -14,7 +14,6 @@ import java.security.spec.X509EncodedKeySpec;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -371,7 +370,7 @@ public class GameProfile {
      * The model used for a profile texture.
      */
     public enum TextureModel {
-        NORMAL,
+        WIDE,
         SLIM;
     }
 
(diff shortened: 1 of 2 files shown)


AlexProgrammerDE:
Mojang AuthLib has similar code:
image

AlexProgrammerDE:
Also fixes now #837

protocol/src/main/java/org/geysermc/mcprotocollib/auth/GameProfile.java
```diff
@@ -390,7 +389,7 @@ public static class Texture {
          */
         public Texture(String url, Map<String, String> metadata) {
             this.url = url;
-            this.metadata = new HashMap<>(metadata);
+            this.metadata = metadata;
```
onebeastchris, line 392: Why was this changed? Is the metadata supposed to be mutable now? (Assuming that was the purpose of the copy)

AlexProgrammerDE: IntelliJ would claim metadata is never null while it can be null due to gson. Idk how gson does it, but it's able to set the metadata field to null.

onebeastchris: Interesting - is null metadata valid?

AlexProgrammerDE: yeah when a skin is wide, metadata field is missing from textures, if it's slim the metadata field is present

onebeastchris: <https://github.com/tinytengu/minecraft-authlib/blob/main/src/main/java/com/mojang/authlib/minecraft/MinecraftProfileTexture.java> Got curious and checked, seems to be what minecraft's authlib is doing, so that looks fine.

onebeastchris (approved):
lgtm

PR #841, opened 2024-07-20
Split incoming/outgoing packet registry, transition protocol states correctly

AlexProgrammerDE:
These changes are necessary to avoid race-conditions in the code and to prepare the codebase for ViaVersion support.


AlexProgrammerDE:
PR is ready for review

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -258,7 +258,17 @@ public interface Session {
...
+        this.send(packet, null);
+    }
+
+    /**
+     * Sends a packet and runs the specified callback when the packet has been sent.
+     *
+     * @param packet Packet to send.
+     * @param onSent Callback to run when the packet has been sent.
+     */
+    void send(Packet packet, Runnable onSent);
```
onebeastchris: Let's add nonnull/nullable annotations here

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -301,4 +311,23 @@ default void disconnect(@NonNull Component reason) {
      * @param cause Throwable responsible for disconnecting.
      */
     void disconnect(@NonNull Component reason, @Nullable Throwable cause);
+
+    /**
+     * Auto read in netty means that the server is automatically reading from the channel.
+     * Turning it off means that we won't get more packets being decoded unless we call read() on the channel.
+     * We use this to hold off on reading packets until we are ready to process them.
+     * Which is for example when we change the protocol state to
```
onebeastchris: ?
seems incomplete

protocol/src/main/java/org/geysermc/mcprotocollib/network/packet/Packet.java
```diff
@@ -17,4 +17,14 @@ public interface Packet {
     default boolean isPriority() {
         return false;
     }
+
+    /**
+     * Whether the packet is terminal. That means this should be the last packet sent inside a protocol state.
+     * Subsequently, we should disable auto-read when we receive a terminal packet and let the code re-enable auto-read when it's ready to receive more packets.
+     *
+     * @return Whether the packet is terminal.
```
Konicai:
```suggestion
     * Returns whether the packet is terminal. If true, this should be the last packet sent inside a protocol state.
     * Subsequently, {@link Session#setAutoRead(boolean)} should be disabled when a terminal packet is received, until the session has switched into a new state and is ready to receive more packets.
     *
     * @return Whether the packet is terminal.
```
probably messed up the link formatting, and requires import

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
Konicai: Nullable annotation on `onSent`

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -301,4 +311,23 @@ default void disconnect(@NonNull Component reason) {
...
+     *
```
Konicai: last sentence is cutoff

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -301,4 +311,23 @@ default void disconnect(@NonNull Component reason) {
...
+     * @param autoRead Whether to enable auto read or not.
+     *                 Default is true.
+     */
+    void setAutoRead(boolean autoRead);
+
+    default void switchInboundProtocol(Runnable switcher) {
```
Konicai: javadoc, and i'd suggest `switchInboundState` because thats the format the getters/setters use

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpFlowControlHandler.java
```diff
@@ -0,0 +1,13 @@
...
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.flow.FlowControlHandler;
+
+public class TcpFlowControlHandler extends FlowControlHandler {
+    @Override
+    public void read(ChannelHandlerContext ctx) throws Exception {
+        if (ctx.channel().config().isAutoRead()) {
+            super.read(ctx);
+        }
```
Konicai, line 18: comment about why this is necessary

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ClientListener.java
```diff
@@ -148,38 +153,28 @@ public void packetReceived(Session session, Packet packet) {
...
+            default -> throw new IllegalStateException("Unexpected value: " + targetState);
+        });
 
-            if (this.targetState == ProtocolState.LOGIN) {
+        switch (this.targetState) {
+            case LOGIN -> {
+                session.switchInboundProtocol(() -> protocol.setInboundState(ProtocolState.LOGIN));
+                session.send(intention, () -> protocol.setOutboundState(ProtocolState.LOGIN));
                 GameProfile profile = session.getFlag(MinecraftConstants.PROFILE_KEY);
                 session.send(new ServerboundHelloPacket(profile.getName(), profile.getId()));
```
Konicai: looks like the inboundState switch and intention send can be pulled out of both cases, out of the switch

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -172,99 +168,118 @@ public void packetReceived(Session session, Packet packet) {
...
+                        session.setFlag(MinecraftConstants.PING_KEY, System.currentTimeMillis() - this.keepAliveTime);
+                    } else {
+                        session.disconnect(Component.translatable("disconnect.timeout"));
+                    }
                 }
             } else if (packet instanceof ServerboundConfigurationAcknowledgedPacket) {
-                protocol.setState(ProtocolState.CONFIGURATION);
+                // The developer who sends ClientboundStartConfigurationPacket needs to setOutboundState to CONFIGURATION
+                // after sending the packet. We can't do it in this class because it needs to be a method call right after it was sent.
+                // Using nettys event loop to change outgoing state may cause differences to vanilla.
```
Konicai, line 183: can we not check for `ClientboundStartConfigurationPacket` in `packetSent`, and setOutboundState to CONFIGURATION there? `packetSent` should be invoked more or less right after the `onSent` callback would be?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -172,99 +168,118 @@ public void packetReceived(Session session, Packet packet) {
...
+                protocol.setOutboundState(ProtocolState.GAME);
+                session.switchInboundProtocol(() -> protocol.setInboundState(ProtocolState.GAME));
                 ServerLoginHandler handler = session.getFlag(MinecraftConstants.SERVER_LOGIN_HANDLER_KEY);
                 if (handler != null) {
                     handler.loggedIn(session);
                 }
 
                 if (session.getFlag(MinecraftConstants.AUTOMATIC_KEEP_ALIVE_MANAGEMENT, true)) {
-                    new Thread(new KeepAliveTask(session)).start();
+                    new Thread(() -> keepAlive(session)).start();
```
Konicai: are we gonna end up with an amount of threads equal to the how many configuration phases we've gone through?

Konicai (dismissed):
nice, especially like the cleanup in the listeners

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: The "proper" way to handle keep-alive is to reset the keepalive listener of the session during protocol switching.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: I'd rather have the developers explicitly switch them like in vanilla. The entire onSent thing is made because of that reason. Stuff in packetSent may be not be handled immediately because packets sometimes get handled on the event loop. However onSent is guaranteed to be handled inside the netty pipeline.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: Maybe I should rework this a bit

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: ClientboundStartConfigurationPacket is not isPriority
therefore it gets handled on the event loop 
so if you 
send(ClientboundStartConfigurationPacket)
and
send(SomeConfigurationPacket)
as a developer right after each other, it'll error because the event loop is having a race condition with the netty pipeline 
One other fix would be setting ClientboundStartConfigurationPacket as isPriority, but I don't wanna touch that with this PR
onSent allows the dev to change the state the next packets gets decoded with immediately without the chance of race conditions because of off-pipeline processing

Konicai: I'm sorry if I don't understand.

> Stuff in packetSent may be not be handled immediately because packets sometimes get handled on the event loop. However onSent is guaranteed to be handled inside the netty pipeline.

I read them as being called right after eachother? https://github.com/AlexProgrammerDE/MCProtocolLib/blob/2b310d2618f23127640a85194623126b1eef594b/protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java#L271-L275

Also, isPriority doesn't seem to apply to outbound packets?

AlexProgrammerDE: Ah yeah you're right. I was thinking too much from the clients perspective that all packets are in the event loop. Would still be weird though to add back packetSent... 
It makes it harder to compare logic with vanilla.

Konicai: It does have the distinction now that any packets handled by ServerListener in the packetSent event are packets that have been sent by the library user, not internal MCPL. I think its worth it to not introduce state switching to the library user for a single case (if i understand correctly)

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -301,4 +311,29 @@ default void disconnect(@NonNull Component reason) {
...
      */
     void disconnect(@NonNull Component reason, @Nullable Throwable cause);
+
+    /**
+     * Auto read in netty means that the server is automatically reading from the channel.
+     * Turning it off means that we won't get more packets being decoded until we turn it back on.
+     * We use this to hold off on reading packets until we are ready to process them.
+     * For example this is used for switching inbound states with {@link #switchInboundState(Runnable)}.
+     *
+     * @param autoRead Whether to enable auto read or not.
```
onebeastchris:
```suggestion
     * @param autoRead Whether to enable auto read.
```

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCodec.java
```diff
@@ -59,13 +72,19 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buf, List<Object> out)
...
 
             if (buf.readableBytes() > 0) {
                 throw new IllegalStateException("Packet \"" + packet.getClass().getSimpleName() + "\" not fully read.");
             }
 
             out.add(packet);
+
+            if (log.isDebugEnabled()) {
+                log.debug("Decoded packet: {}", packet.getClass().getSimpleName());
+            }
```
onebeastchris: I wonder if we should just debug print a packet once, otherwise there's 2 debug lines for every packet.. might be a bit overkill 

If the decoding fails, it would print an error with the packet id anyways

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
```diff
@@ -186,17 +201,35 @@ protected PacketEncryption enableEncryption(Key key) {
...
+     * Gets the current outbound {@link ProtocolState} we're in.
+     *
+     * @return The current outbound {@link ProtocolState}.
+     */
+    public ProtocolState getOutboundState() {
+        return this.outboundState;
+    }
+
+    public void setInboundState(ProtocolState state) {
+        log.debug("Setting inbound state to: {}", state);
```
onebeastchris:
```suggestion
        log.debug("Setting inbound protocol state to: {}", state);
```
Might be just a bit clearer

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
```diff
@@ -186,17 +201,35 @@ protected PacketEncryption enableEncryption(Key key) {
...
+
+        this.inboundState = state;
+        this.inboundStateRegistry = this.codec.getCodec(state);
+    }
+
+    public void setOutboundState(ProtocolState state) {
+        log.debug("Setting outbound state to: {}", state);
```
onebeastchris:
```suggestion
        log.debug("Setting outbound protocol state to: {}", state);
```
Same as above

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -172,100 +172,136 @@ public void packetReceived(Session session, Packet packet) {
...
-                } catch (InterruptedException e) {
-                    break;
+                    currentKeepAliveState.keepAlivePending = true;
+                    currentKeepAliveState.keepAliveChallenge = time;
+                    currentKeepAliveState.keepAliveTime = time;
+                    session.send(new ClientboundKeepAlivePacket(currentKeepAliveState.keepAliveChallenge));
                 }
             }
+
+            // TODO: Implement proper tick loop rather than sleeping
```
onebeastchris, line 291: Is this a TODO for this PR?

onebeastchris (approved):

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: No, not part of this PR.

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
AlexProgrammerDE: Was fixed 3 weeks ago.

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
AlexProgrammerDE: Was fixed 3 weeks ago.

onebeastchris: Packet is still not marked, which I'd make NonNull

AlexProgrammerDE: done

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: This is now not even possible anymore since explicit flushes after the writeAndFlush are needed due to race conditions. The code doesn't use onSent anymore for state switching. Developers need to manually switch state after sending the packet.

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCodec.java
AlexProgrammerDE: True, but imo it should be kept like this in case of for example deadlocks. If the deserialization never returns a result or an error and instead infitely blocks the thread we'd have no idea what is going on.

onebeastchris: If it returns an error, that would be logged already, no? 

My concern would be flooding debug logs with double the lines for just a little more information that is, more or less, redundant

Camotoy: If we really want this, can we set one to `fine`/`trace` logging level?

AlexProgrammerDE: Yeah ig trace works

Redned235 (approved):

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCodec.java
AlexProgrammerDE: Done

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
```diff
@@ -115,9 +115,10 @@ public void initChannel(Channel channel) {
 
                     pipeline.addLast("sizer", new TcpPacketSizer(protocol.getPacketHeader(), getCodecHelper()));
 
-                    pipeline.addLast("codec", new TcpPacketCodec(TcpClientSession.this, true));
-                    pipeline.addLast("manager", TcpClientSession.this);
-                }
+                    pipeline.addLast("flow-control", new TcpFlowControlHandler());
+                            pipeline.addLast("codec", new TcpPacketCodec(TcpClientSession.this, true));
+                            pipeline.addLast("manager", TcpClientSession.this);
```
onebeastchris:
```suggestion
                    pipeline.addLast("flow-control", new TcpFlowControlHandler());
                    pipeline.addLast("codec", new TcpPacketCodec(TcpClientSession.this, true));
                    pipeline.addLast("manager", TcpClientSession.this);
```

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpSession.java
```diff
@@ -221,18 +221,27 @@ public boolean isConnected() {
...
-    public void send(Packet packet) {
+    public void send(@NonNull Packet packet, @Nullable Runnable onSent) {
         if (this.channel == null) {
             return;
         }
 
+        if (!this.channel.eventLoop().inEventLoop()) {
+            this.channel.eventLoop().execute(() -> this.send(packet, onSent));
+            return;
+        }
```
onebeastchris, line 228: comment please about this being java client behavior

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ClientListener.java
```diff
@@ -132,9 +136,13 @@ public void packetReceived(Session session, Packet packet) {
                     newSession.connect(true, true);
                 }
             }
-        } else if (protocol.getState() == ProtocolState.CONFIGURATION) {
-            if (packet instanceof ClientboundFinishConfigurationPacket) {
+        } else if (protocol.getInboundState() == ProtocolState.CONFIGURATION) {
+            if (packet instanceof ClientboundKeepAlivePacket keepAlivePacket && session.getFlag(MinecraftConstants.AUTOMATIC_KEEP_ALIVE_MANAGEMENT, true)) {
+                session.send(new ServerboundKeepAlivePacket(keepAlivePacket.getPingId()));
```
onebeastchris, line 146: nice fix, should resolve issues with polymer

onebeastchris (approved):
lgtm

AlexProgrammerDE:
Race conditions and weird disconnects are fixed. This PR can be merged.

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
```diff
@@ -115,9 +115,12 @@ public void initChannel(Channel channel) {
 
                     pipeline.addLast("sizer", new TcpPacketSizer(protocol.getPacketHeader(), getCodecHelper()));
 
+                    pipeline.addLast("flow-control", new TcpFlowControlHandler());
+
                     pipeline.addLast("codec", new TcpPacketCodec(TcpClientSession.this, true));
+                    pipeline.addLast("packet-filter", new DropPacketFilter());
                     pipeline.addLast("manager", TcpClientSession.this);
-                }
+}
```
onebeastchris: ?

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpServer.java
```diff
@@ -66,7 +66,10 @@ public void initChannel(Channel channel) {
 
                 pipeline.addLast("sizer", new TcpPacketSizer(protocol.getPacketHeader(), session.getCodecHelper()));
 
+                pipeline.addLast("flow-control", new TcpFlowControlHandler());
+
```
onebeastchris:
```suggestion
```

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
```diff
@@ -115,9 +115,12 @@ public void initChannel(Channel channel) {
 
                     pipeline.addLast("sizer", new TcpPacketSizer(protocol.getPacketHeader(), getCodecHelper()));
 
+                    pipeline.addLast("flow-control", new TcpFlowControlHandler());
+
```
onebeastchris:
```suggestion
```

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCodec.java
```diff
@@ -23,57 +30,87 @@ public TcpPacketCodec(Session session, boolean client) {
 
     @SuppressWarnings({"rawtypes", "unchecked"})
     @Override
-    public void encode(ChannelHandlerContext ctx, Packet packet, ByteBuf buf) {
-        int initial = buf.writerIndex();
+    public void encode(ChannelHandlerContext ctx, Packet packet, List<Object> out) {
+        if (log.isTraceEnabled()) {
+            log.trace("Encoding packet: {}", packet.getClass().getSimpleName());
+        }
```
onebeastchris: do we need the trace enabled check? doesn't log.trace() perform it too before formatting the string?

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/DropPacketFilter.java
```diff
@@ -0,0 +1,23 @@
+package org.geysermc.mcprotocollib.network.tcp;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelOutboundHandlerAdapter;
+import io.netty.channel.ChannelPromise;
+
+public class DropPacketFilter extends ChannelOutboundHandlerAdapter {
```
onebeastchris: I understand why it's named filter, but later in the pipeline it's referred to as "packet_filter" which imo doesn't quite convey the functionality of the pipeline "flushing" (why'd you rename it to dropping anyways? Just curious)

Maybe `FlushHandler` or something?

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpPacketCodec.java
AlexProgrammerDE: This is to avoid array allocations and expensive method calls, packet.getClass().getSimpleName() will lag a tiny bit and the trace enabled check prevents it being called.

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/DropPacketFilter.java
AlexProgrammerDE: Yeah idk about the inconsistency, I have renamed and replaced this thing like 5 times until I got this working impl. I'll rename it to FlushHandler.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
Konicai: > Would still be weird though to add back packetSent...
It makes it harder to compare logic with vanilla.

Minecraft is also the maintainer AND user of its network code—I think it is worth it here for the sake of ease of use.

Minecraft has a method for sending ClientboundStartConfigurationPacket and switching outbound protocol, but I'm not sure this project's structure would facilitate something like that?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -172,100 +172,136 @@ public void packetReceived(Session session, Packet packet) {
             } else if (packet instanceof ServerboundPingRequestPacket pingRequestPacket) {
                 session.send(new ClientboundPongResponsePacket(pingRequestPacket.getPingTime()));
             }
-        } else if (protocol.getState() == ProtocolState.GAME) {
+        } else if (protocol.getInboundState() == ProtocolState.GAME) {
             if (packet instanceof ServerboundKeepAlivePacket keepAlivePacket) {
-                if (keepAlivePacket.getPingId() == this.lastPingId) {
-                    long time = System.currentTimeMillis() - this.lastPingTime;
-                    session.setFlag(MinecraftConstants.PING_KEY, time);
+                KeepAliveState currentKeepAliveState = this.keepAliveState;
```
Konicai: What specifically is this being cached to a variable for? The field value seems to be mutated when updating the keep alive, but replaced on login and config start/end.

If we are trying to prevent races with the keep alive thread, have we considered making `KeepAliveState` immutable?

Also, could `keepAlivePending` be replaced with the field being null when not pending, and have the field be volatile?

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
```diff
@@ -172,100 +172,136 @@ public void packetReceived(Session session, Packet packet) {
...
+            if (packet instanceof ServerboundKeepAlivePacket keepAlivePacket) {
+                KeepAliveState currentKeepAliveState = this.keepAliveState;
+                if (currentKeepAliveState != null) {
+                    if (currentKeepAliveState.keepAlivePending && keepAlivePacket.getPingId() == currentKeepAliveState.keepAliveChallenge) {
+                        currentKeepAliveState.keepAlivePending = false;
+                        session.setFlag(MinecraftConstants.PING_KEY, System.currentTimeMillis() - currentKeepAliveState.keepAliveTime);
+                    } else {
+                        session.disconnect(Component.translatable("disconnect.timeout"));
+                    }
+                }
```
Konicai: This code is fairly long and shared between the two states, would be good to move to a method

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: This is coded closer to mojangs impl.
The pending field and other fields are part of the commonpacketlistener. And each protocol state keeps track of separate keepalives.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: I'd target a rewrite that more closely resembles mojangs impl where each protocol state listener extends a common listener with common logic. I'd not worry too much about duplicated code rn.

AlexProgrammerDE: done

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
```diff
@@ -84,7 +90,8 @@ public MinecraftProtocol(PacketCodec codec) {
         this.codec = codec;
         this.targetState = ProtocolState.STATUS;
 
-        this.setState(ProtocolState.HANDSHAKE);
+        this.setInboundState(ProtocolState.HANDSHAKE);
```
Camotoy: Seems like we could just have a setState/setStates method just for convenience.

protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
```diff
@@ -100,7 +101,7 @@ public void connect(boolean wait, boolean transferring) {
             .localAddress(bindAddress, bindPort)
             .handler(new ChannelInitializer<>() {
                 @Override
-                public void initChannel(Channel channel) {
+                public void initChannel(@NotNull Channel channel) {
```
onebeastchris: Please use the NonNull annotation from the checkerframework lib

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
AlexProgrammerDE: Is this really necessary? The only real reason for the protocol to be set for both at once is for handshake. There's currently no other reason for setting both inbound and outbound at once. There are also separate operations on the Session used for switching protocols.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: Changing it now would make diffing it with vanilla harder. Also, if I restructure it, there's the possibility something I code doesn't properly match up with vanilla.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ServerListener.java
AlexProgrammerDE: Could you elaborate? 
What I'm saying is that the calling thread for the switchOutbound matters. Always forcing it to be inside netty threads could lead to deadlocks if the event loop is used for the inbound processing.

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
Camotoy: It's done three times here. Might as well have a private method for it?

AlexProgrammerDE: We could have a private method to reset current state to handshake. Will add that tomorrow.

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -255,4 +267,48 @@ default void disconnect(@NonNull Component reason) {
...
+     *                 Default is true.
+     */
+    void setAutoRead(boolean autoRead);
+
+    /**
+     * Returns the underlying netty channel of this session.
+     *
+     * @return The netty channel
+     */
+    Channel getChannel();
```
Konicai, line 287: this doesn't need to be exposed if `switchOutboundState` is implemented in TcpSession

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
```diff
@@ -255,4 +267,48 @@ default void disconnect(@NonNull Component reason) {
...
+        switcher.run();
+
+        // We switched to the new inbound state
+        // we can start reading again
+        setAutoRead(true);
+    }
+
+    /**
+     * Flushes all packets due to be sent and changes the outbound state of the session.
+     * This makes sure no other threads scheduled packets to be sent.
```
Konicai: grammar

protocol/src/main/java/org/geysermc/mcprotocollib/network/Session.java
AlexProgrammerDE: Yes, this is new API. It's already public for TcpSession. The reason why I'm exposing it is so it's easier to make ViaVersion integrations in the future as it needs direct channel access.

Konicai: sounds good 👍

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/MinecraftProtocol.java
AlexProgrammerDE: done

onebeastchris:
As discussed on discord, being able to disable/suppress the packet en/decoding logging while in debug mode using e.g. markers as flags would be ideal.
Alternatively, the logging addition could get implemented in a separate PR to not block this PR further - IMO it's the only thing remaining

AlexProgrammerDE:
Should be able to be reviewed now. Added the markers.

PR #846, opened 2024-08-05
Update TcpClientSession.java

waterghost-2046:
The code has been modified to prevent SRV record lookup when the host is in the format ip(*.*.*.*) This is achieved by adding an additional condition !this.host.matches("\\*\\.\\*\\.\\*\\.\\*") to the if statement that checks whether SRV record resolution should be attempted.

### DIFF
--- a/protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
+++ b/protocol/src/main/java/org/geysermc/mcprotocollib/network/tcp/TcpClientSession.java
@@ -152,7 +152,7 @@ public class TcpClientSession extends TcpSession {
         String name = this.getPacketProtocol().getSRVRecordPrefix() + "._tcp." + this.getHost();
         log.debug("Attempting SRV lookup for \"{}\".", name);
 
-        if (getFlag(BuiltinFlags.ATTEMPT_SRV_RESOLVE, true) && (!this.host.matches(IP_REGEX) && !this.host.equalsIgnoreCase("localhost") && !this.host.matches("\\*\\.\\*\\.\\*\\.\\*"))) {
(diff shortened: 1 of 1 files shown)


AlexProgrammerDE:
Isn't this similar to what IP_REGEX already does?
Also, since you match wildcards, wouldn't also domains like a.b.example.com never be looked up? It didn't seem like it only matches digits at a first glance.

AlexProgrammerDE:
Maybe we should update the system to use 1:1 the same technique vanila uses to avoid differences to vanilla.

waterghost-2046:
Right? If you look at the logs below, the SRV is being queried from the IP address.

```
00:42:52.816 [pool-2-thread-225] DEBUG org.geysermc.mcprotocollib.network.tcp.TcpClientSession -- Attempting SRV lookup for "_minecraft._tcp.146.56.180.111".
        at org.geysermc.mcprotocollib.network.tcp.TcpClientSession.connect(TcpClientSession.java:90)
        at org.geysermc.mcprotocollib.network.tcp.TcpSession.connect(TcpSession.java:76)
        at org.geysermc.mcprotocollib.network.tcp.TcpSession.connect(TcpSession.java:71)
        at forgeban.xyz.Main.lambda$main$2(Main.java:88)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
00:42:52.816 [pool-2-thread-634] DEBUG org.geysermc.mcprotocollib.network.tcp.TcpClientSession -- Attempting SRV lookup for "_minecraft._tcp.146.56.180.111".
```

PR #859, opened 2024-09-10
Allow accessing backing class for a channel instead of only its factory

AlexProgrammerDE:
Some netty libraries that are poorly coded only accept classes. Example: https://github.com/CloudburstMC/Network/pull/42


Camotoy:
How would you use this for your MCProtocolLib projects?

AlexProgrammerDE:
@Camotoy Like this: https://github.com/AlexProgrammerDE/SoulFire/blob/b629ea854ca35cd3b0fee8528ecff12477bda6cd/server/src/main/java/com/soulfiremc/server/protocol/netty/ViaClientSession.java#L127

onebeastchris (approved):

Konicai (approved):

PR #862, opened 2024-10-10
Fix/proxy protocol

AlexProgrammerDE:
[Fix haproxy not having a working remote address](https://github.com/GeyserMC/MCProtocolLib/commit/187a306b867f99df33d20ff9d62b0fbdadfebd96)

Netty resolving is removed, if we can't resolve an address, netty can't either. The main difference is that netty may resolve addresses using an incorrect dns server/proxy unlike our code where we keep full control of the dns server and other things used to resolve addresses like users want us to resolve them.


AlexProgrammerDE:
nvm about the netty resolving stuff, the PR now uses the old code for haproxy and just cleans up a finally block

onebeastchris (approved):

PR #863, opened 2024-10-11
Move reusable methods to a separate helper class

AlexProgrammerDE:
This way we allow other apps such as Geyser LocalSession to use these currently private methods without needing to copy over the code.


onebeastchris (approved):
lgtm

PR #865, opened 2024-10-21
Improve MCPL ticking data and behaviour

AlexProgrammerDE:
Goal of this PR is to get EXACT 1:1 ticking behavior with vanilla. Developers will now be able to provide their own async packet tick executors rather than mcpl forcing it to be an eventloop.
Some packets have been also put into proper packages. Such as the cookie packets in the cookies package and ping packets in the ping package, just like vanilla.
Some packets have exceptions and are only after specific code/in specific state ticked async. Those have been marked with a // GAME THREAD DETAIL comment.
This PR is ready for merging, thanks.


AlexProgrammerDE:
This PR is now updated for 1.21.3

protocol/src/main/java/org/geysermc/mcprotocollib/network/packet/Packet.java
```diff
@@ -9,13 +9,14 @@
...
+     * Gets whether the packet should run on an async game thread rather than blocking the network (Netty) thread.
+     * Packets that qualify for this are usually packets with an ensureRunningOnSameThread call at the top of their packet listener method in the Minecraft code.
+     * Packets which need extra attention because they aren't "fully" handled async are marked using.
+     * // GAME THREAD DETAIL comments in the MCProtocolLib code.
      *
-     * @return Whether the packet has priority.
+     * @return Whether the packet be handled async from the Netty thread.
      */
-    default boolean isPriority() {
+    default boolean shouldRunOnGameThread() {
```
onebeastchris, line 19: Wouldn't it make more sense to flip this to instead reflect if it shouldn't run on the game thread?
Based on the diff, it looks like most things run on the game thread, instead of the other way around

AlexProgrammerDE: This is so it is easier to read together with mojang. Because when you use netty all things run on the netty thread by default. Minecraft explicitly marks every packet as ticked. The reaspn is that you could accidentally implicitly tick a packet that shouldn't be ticked.

AlexProgrammerDE: Ticking that packet that shouldn't be ticked is more dangerous than a non-ticked packet. Especially when it comes to protocols, order and state switching.

onebeastchris: In that case, shouldn't the method have no default body? Just to ensure that it's not "forgotten" but set on purpose?

AlexProgrammerDE: Yeah I just didn't wanna edit more classes ig lol. Would you like me to explicitly set it for each packet?

AlexProgrammerDE: Resolved in private chat, we're keeping this impl

AlexProgrammerDE:
PR is now fully ready for review & merge

onebeastchris (approved):

protocol/src/main/java/org/geysermc/mcprotocollib/network/packet/Packet.java
```diff
@@ -9,13 +9,14 @@
 public interface Packet {
 
     /**
-     * Gets whether the packet has handling priority.
-     * If the result is true, the packet will be handled immediately after being
-     * decoded.
+     * Gets whether the packet should run on an async game thread rather than blocking the network (Netty) thread.
+     * Packets that qualify for this are usually packets with an ensureRunningOnSameThread call at the top of their packet listener method in the Minecraft code.
+     * Packets which need extra attention because they aren't "fully" handled async are marked using.
```
onebeastchris:
```suggestion
     * Packets which need extra attention because they aren't "fully" handled async are marked using
```

PR #869, opened 2024-11-29
Reimplement session structure

AlexProgrammerDE:
Benefits are clear, developers can easily provide custom behaviour by overriding the server and client classes to provide custom channels (like geyser does for localchannel) and also custom channel handlers can be provided, like for ViaVersion support.

Transfers were moved to a flag. There are now ClientSession and ServerSession interfaces which extend Session. This way a ServerSession doesn't have connect methods. Binding to UnixDomainSockets or any other type of SocketAddress is now supported and the client can connect to those too. Tcp was renamed to Net because now non-tcp protocols like local channels and udp (for bedrock ViaBedrock) can be used.


protocol/src/main/java/org/geysermc/mcprotocollib/network/net/NetClientSession.java
```diff
@@ -0,0 +1,154 @@
...
+    public NetClientSession(SocketAddress remoteAddress, SocketAddress bindAddress, PacketProtocol protocol, ProxyInfo proxy) {
+        this(remoteAddress, bindAddress, protocol, proxy, DefaultPacketHandlerExecutor.createExecutor());
+    }
+
+    public NetClientSession(SocketAddress remoteAddress, SocketAddress bindAddress, PacketProtocol protocol, ProxyInfo proxy, Executor packetHandlerExecutor) {
+        super(remoteAddress, protocol, packetHandlerExecutor);
+        this.bindAddress = bindAddress;
+        this.proxy = proxy;
+        this.codecHelper = protocol.createHelper();
+    }
```
onebeastchris: Thoughts on making these a builder with default values instead of multiple constructors?

AlexProgrammerDE:
Implementation note: the MinecraftConstants.CLIENT_HOST and CLIENT_PORT flags are dynamic flags used by the ClientSessionListener. If not explicitly set, they'll cast the remote address to an InetSocketAddress to take the values. However for developers using anything other like UnixDomainSockeAddresst or LocalChannelAddress they'll have to set those flags to avoid an cast error.

example/src/main/java/org/geysermc/mcprotocollib/network/example/PingServerTest.java
```diff
@@ -25,11 +26,11 @@ public static void main(String[] args) {
             return;
         }
 
-        Server server = new TcpServer("127.0.0.1", 25565, TestProtocol::new);
+        Server server = new NetServer(new InetSocketAddress("127.0.0.1", 25565), TestProtocol::new);
```
onebeastchris: Not sure I like the prefix of `Net`, it's very generic (?) But, not a deal breaker imo

Just an idea - maybe we could call it `JavServer` and `JavaClientSession`? It would deviate from the previous intention, but, it would be clearer

protocol/src/main/java/org/geysermc/mcprotocollib/network/helper/NettyHelper.java
```diff
@@ -102,12 +104,33 @@ public static void initializeHAProxySupport(Session session, Channel channel) {
...
+                        case TCP4 -> {
+                            destinationAddress = "0.0.0.0";
+                            destinationPort = 0;
+                        }
+                        case TCP6 -> {
+                            destinationAddress = "0:0:0:0:0:0:0:0";
+                            destinationPort = 0;
+                        }
+                        default -> throw new UnsupportedOperationException("Unsupported proxied protocol: " + proxiedProtocol);
+                    };
```
onebeastchris: just in case this ever causes issues, let's a debug message stating what the destination address has been set to?

protocol/src/main/java/org/geysermc/mcprotocollib/network/net/NetClientSession.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    @Override
+    public PacketCodecHelper getCodecHelper() {
+        return this.codecHelper;
+    }
+
+    protected EventLoopGroup getEventLoopGroup() {
+        createEventLoopGroup();
```
onebeastchris: any particular reason the event group null check was moved to the method instead of being here?

protocol/src/main/java/org/geysermc/mcprotocollib/network/net/NetClientSession.java
```diff
@@ -0,0 +1,153 @@
...
+        return new MinecraftChannelInitializer<>(channel -> {
+            PacketProtocol protocol = getPacketProtocol();
+            protocol.newClientSession(NetClientSession.this);
+
+            return NetClientSession.this;
+        }, true) {
+            @Override
+            public void initChannel(@NonNull Channel channel) throws Exception {
+                NettyHelper.addProxy(proxy, channel.pipeline());
+
```
onebeastchris:
```suggestion
```

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ClientListener.java
```diff
@@ -167,16 +170,21 @@ public void packetReceived(Session session, Packet packet) {
...
         Session session = event.getSession();
         MinecraftProtocol protocol = (MinecraftProtocol) session.getPacketProtocol();
-        ClientIntentionPacket intention = new ClientIntentionPacket(protocol.getCodec().getProtocolVersion(), session.getHost(), session.getPort(), switch (targetState) {
-            case LOGIN -> transferring ? HandshakeIntent.TRANSFER : HandshakeIntent.LOGIN;
-            case STATUS -> HandshakeIntent.STATUS;
-            default -> throw new IllegalStateException("Unexpected value: " + targetState);
-        });
+        ClientIntentionPacket intention = new ClientIntentionPacket(
+            protocol.getCodec().getProtocolVersion(),
+            session.getFlagSupplied(MinecraftConstants.CLIENT_HOST, () -> ((InetSocketAddress)session.getRemoteAddress()).getHostString()),
```
onebeastchris:
```suggestion
            session.getFlagSupplied(MinecraftConstants.CLIENT_HOST, () -> ((InetSocketAddress) session.getRemoteAddress()).getHostString()),
```

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/ClientListener.java
```diff
@@ -167,16 +170,21 @@ public void packetReceived(Session session, Packet packet) {
...
+            session.getFlagSupplied(MinecraftConstants.CLIENT_PORT, () -> ((InetSocketAddress)session.getRemoteAddress()).getPort()),
```
onebeastchris:
```suggestion
            session.getFlagSupplied(MinecraftConstants.CLIENT_PORT, () -> ((InetSocketAddress) session.getRemoteAddress()).getPort()),
```

onebeastchris (commented):
Other than these RFCs/minor touches, lgtm

example/src/main/java/org/geysermc/mcprotocollib/network/example/PingServerTest.java
AlexProgrammerDE: Naming was inspired by NetMinecraft. https://github.com/RaphiMC/NetMinecraft/blob/main/core/src/main/java/net/raphimc/netminecraft/netty/connection/NetClient.java
The reason it's not JavaServer is that Java implies tcp. However in the future a "JavaClient" could also connect to bedrock servers, which would be confusing. (Using the ViaVersion addon [ViaBedrock](https://github.com/RaphiMC/ViaBedrock))

protocol/src/main/java/org/geysermc/mcprotocollib/network/net/NetClientSession.java
AlexProgrammerDE: Better isolates EVENT_LOOP_GROUP to only be referenced from one method. Also imo makes this method more readable for devs looking to override this.

Redned235 (changes requested):
Code generally looks fine fine to me, but the main nitpick I have is with the naming. `net` is just an abbreviation of `network`, and it doesn't make the most sense IMO to have a second network package nested inside of `network` already. I'd suggest renaming the `net` package to `session` to be a bit more clear to the end user.

Additionally I'd suggest the following names:
- `NetSession` -> `NetworkSession`
- `NetClientSession` -> `ClientNetworkSession`
- `NetServerSession` -> `ServerNetworkSession`

I'd drop the `Net` prefix for all the other remaining classes aside from `NetFlowControlHandler` being renamed to `AutoReadFlowControlHandler` or retaining the previous `Tcp` prefix as the doc indicates this is still for TCP connections. May be better for these classes to be placed in a `netty` package as well.

AlexProgrammerDE:
Alright implemented Redneds advice

Redned235 (approved):

protocol/src/main/java/org/geysermc/mcprotocollib/network/net/NetClientSession.java
AlexProgrammerDE: This is done now. A new ClientNetworkSessionFactory class was added.

onebeastchris (approved):

PR #878, opened 2025-01-18
Staticify all serialization helpers

AlexProgrammerDE:
First of all, why am I changing the whole API so that MinecraftCodecHelper is now static instead of an instance? The reason is that these methods are better used statically. Right now the biggest issue with MCPL I see is its reliance on "serialization instances" that are being passed around everywhere. Why were those even made in the first place? Because there was a belief a few years ago that MCPL could somehow ever support multiple versions of Minecraft. The idea is good however it is unrealistic. Packets change heavily between versions and so do more complex serialization types. Serialization instances were made with the idea somehow you could implement multiple codechelpers for each version and "downgrade" packets and bind them together using UnsupportedOperationException. In my opinion this is not a route MCPL should be taking. Instead my plan is to add a module that allows ViaVersion to be used with MCPL to allow the desired multi-version support. Without having to pass serialization instances around, the API becomes a lot simpler. Especially with the removal of the parent class of ItemCodecHelper. (Which previously extended MinecraftCodecHelper for some reason) Now it is clear what you're calling and there are way more lambda/inlining opportunities around the codebase. This makes the API more similar to the approach Velocity and ViaVersion are also taking with their static serialization methods. 

Classes renamed:
- MinecraftCodecHelper -> MinecraftTypes
- ItemCodecHelper -> ItemTypes

Classes removed:
- PacketCodecHelper interface (not needed anymore)
- BasePacketCodec (merged into MinecraftTypes)

Why was BasePacketCodec not renamed to BaseTypes? Because this library is slowly going to more and more specialize on only minecraft, not needing to retain a special implementation for developers who use it just for PacketLib.


basaigh (approved):

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/ingame/clientbound/level/ClientboundGameEventPacket.java
```diff
@@ -25,7 +24,7 @@ public class ClientboundGameEventPacket implements MinecraftPacket {
     private final @NonNull GameEvent notification;
     private final GameEventValue value;
 
-    public ClientboundGameEventPacket(ByteBuf in, MinecraftCodecHelper helper) {
+    public ClientboundGameEventPacket(ByteBuf in) {
         this.notification = GameEvent.from(in.readUnsignedByte());
         float value = in.readFloat();
         // TODO: Handle this in MinecraftCodecHelper
```
onebeastchris: should be removed, as it's now irrelevant

protocol/src/main/java/org/geysermc/mcprotocollib/protocol/packet/ingame/clientbound/level/ClientboundGameEventPacket.java
```diff
@@ -51,7 +50,7 @@ public ClientboundGameEventPacket(ByteBuf in, MinecraftCodecHelper helper) {
     }
 
     @Override
-    public void serialize(ByteBuf out, MinecraftCodecHelper helper) {
+    public void serialize(ByteBuf out) {
         out.writeByte(this.notification.ordinal());
         float value = 0;
         // TODO: Handle this in MinecraftCodecHelper
```
onebeastchris: as above

onebeastchris (approved):
Looks good overall - found 2 comments that should be removed too now that multi-versioning isn't a concern anymore. There might be more comments like them, just found these while skimming

AlexProgrammerDE:
Replaced the three comments there were. Better keep them rather than remove because they can still be moved to MinecraftTypes.

PR #881, opened 2025-01-30
Rewrite some internal code to allow better disconnect handling

AlexProgrammerDE:
This code changes a few things around the codebase:
- DisconnectionDetails were introduced and allow more extension in the future. Those can be used in the future to attach even more info to disconnects. (Like Minecraft does)
- PacketFlow was introduced to handle server disconnect packets inside NetworkSession
- NetworkSession and children now only accept MinecraftProtocol which removes unnecessary casts across the codebase and makes it easier to do Minecraft-related changes specifically. The library slowly moves away from accepting non-minecraft serlization for network sessions.
- Disconnect method behaviour was changed to match Minecraft, may affect events, but tests still pass and disconnects still get handled properly with events.
- Errors in packet sending are now fired on the event loop to ensure processing.
- Server packet disconnect handling was moved to NetworkSession, just like Minecrafts Connection class
- Add/remove checks to match Minecrafts behaviour
- isConnected now no longer considers being disconnected to match Minecrafts behaviour (tests still pass)
- Some small syntax cleanups where code was unnecessary.

### DIFF
--- a/example/src/main/java/org/geysermc/mcprotocollib/network/example/ClientSessionListener.java
+++ b/example/src/main/java/org/geysermc/mcprotocollib/network/example/ClientSessionListener.java
@@ -45,11 +45,11 @@ public class ClientSessionListener extends SessionAdapter {
 
     @Override
     public void disconnecting(DisconnectingEvent event) {
-        log.info("CLIENT Disconnecting: {}", event.getReason());
+        log.info("CLIENT Disconnecting: {}", event.getDetails().reason());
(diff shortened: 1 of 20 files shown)


AlexProgrammerDE:
This PR has lost its scope and has some changes I'd like to heavily revise in the future. Thus closing. Some of MCs packet handling requires a tick system and MCPL is not quite there to support ticking. Maybe in the future.

PR #892, opened 2025-03-30
1.21.5

onebeastchris:


AlexProgrammerDE:
This will not be a very useful release for anyone except Geyser if it's missing data component hashing. Will Geyser offer an item hashing implementation or render this library useless for inventory interactions?

PR #894, opened 2025-05-05
Data components map improvements

eclipseisoffline:
This PR makes a couple of small improvements to the `DataComponents` class:

- Added `contains(DataComponentType<?>)` and `remove(DataComponentType<T>)` methods.
- Make `put(DataComponentType<T>, T)` allow null values.
- Add Javadoc to make usage of null values clear.

In the future, an additional, separate class should probably be created for data component patches.

This PR also makes a small change to `DataComponentTypes`: a `fromKey(Key)` method was added.


onebeastchris (approved):
LGTM - this could probably be improved further in the future, but it's definitely a neat improvement for the time being

PR #896, opened 2025-05-21
Migrate to io_uring that's now part of netty-all

AlexProgrammerDE:
Very recently the io_uring netty transport has left incubator state, the io_uring transport is now part of the main netty codebase and is distributed using the netty-all project.


onebeastchris:
This won't build :p

AlexProgrammerDE:
Gradle wrapper action decided to fail for no reason

AlexProgrammerDE:
Replaced by #897

PR #909, opened 2025-08-29
fix: check fix "click outside not holding" click

huper1:
check fix "click outside not holding" click


basaigh (approved):
Lgtm
Looks like an obvious mistake made in 4dd7416966c41764ac574779e76583290f75bf31 which this corrects.

PR #910, opened 2025-09-13
Properly calculate global BPE and add initial value for data palettes

eclipseisoffline:
This PR re-introduces custom global BPE (bits-per-entry) values for data palettes, however, instead of passing these directly, library users pass the size of the applying registry (block state or biome), and MCPL calculates the global BPE value necessary to store all of the values of that registry.

When creating an empty data palette, MCPL now uses a singleton palette instead of a list palette. This saves a bit of space. Library users also have to specify the initial value of the data palette, instead of this defaulting to whatever state is first set in the palette.

`PaletteType.CHUNK` has also been renamed to `PaletteType.BLOCK_STATE` for clarity. Related variables and parameters have also been renamed.


onebeastchris (approved):
lgtm!

basaigh (approved):
lgtm :)

PR #921, opened 2026-02-17
Fix: Writing equipment slot ordinal instead of network id

onebeastchris:
Finishes up https://github.com/GeyserMC/mcprotocollib/commit/1e8f791c323f9a509117957bb003e1d4e12e1ef9


basaigh (approved):
Lgtm

PR #926, opened 2026-05-11
fix(MinecraftTypes): handle empty `EnumSet` in `readEnumSet`

cattyngmd:
when trying to connect to 1.20.6 using ViaProxy it produces `java.lang.IllegalArgumentException: Collection is empty`


basaigh:
Thanks for the PR! I've made a slight alteration to more closely match vanilla, but wouldn't have noticed it without this.

PR #927, opened 2026-06-20
Fix ServerboundSpectatorActionPacket to match vanilla OPTIONAL_VAR_INT

darkcupid412:
It was serialized as a boolean + VarInt, but vanilla uses `ByteBufCodecs.OPTIONAL_VAR_INT` (a single VarInt: `0` = empty, `n` = `of(n-1)`). The extra leading byte desyncs the stream, so a real 26.2 server rejects the packet and disconnects. No API change.


onebeastchris:
Thanks for the PR, good catch!

References:
https://mcsrc.dev/1/26.2/net/minecraft/network/protocol/game/ServerboundSpectatorActionPacket#L12

https://mcsrc.dev/1/26.2/net/minecraft/network/codec/ByteBufCodecs#L112

onebeastchris (approved):

PR #930, opened 2026-07-07
Don't disconnect clients that send ping requests in PLAY state

amyavi:
I'm assuming this was some sort of copy-paste error, as you are supposed to disconnect them in the STATUS state, not the PLAY state, which is the opposite of what was being done in ServerListener.


basaigh (approved):
Thanks for catching this :)

